PR #51, opened 2025-09-16
Added requirements section + `--ipfs-server` arg to the README.md

bkontur:


README.md
```diff
@@ -52,9 +52,31 @@ Stores arbitrary data on IPFS via the `store` extrinsic, provided that either th
 
 ## Prepare for a production
 
+### Requirements
+
+There are no special requirements for the production runtime (just as the usual [validator/node](https://docs.polkadot.com/infrastructure/running-a-validator/#running-a-validator)), except those related to IPFS support.
```
bkontur, line 57: @dmitry-markin Can you please confirm these requirements (I know I asked you this several times already, but this is for record :D ) and also do we need to run all the nodes/validators with `--ipfs-server`, right?

bkontur: > Also, bitswap support is only available with litep2p network backend, but this is the substrate's default, so I am not sure if we need to state this separately.

Yes, we should highlight this in the README.md, can you please add here the arg/setting for this, e.g. `--network=litep2p`, so no body changes that by accident

README.md
```diff
@@ -52,9 +52,31 @@ Stores arbitrary data on IPFS via the `store` extrinsic, provided that either th
...
+With the current configuration, the maximum storage requirement is estimated as follows:
```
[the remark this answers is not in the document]
serban300: Looks good to me.

PR #56, opened 2025-09-25
Added check-metadata-hash

antkve:


bkontur:
@antkve just please make sure we add this to the BW para and BP para (should be there already)

PR #70, opened 2025-10-08
[WIP] Introduce cargo clippy CI check

karolk91:


.github/workflows/check.yml
```diff
@@ -78,6 +78,32 @@ jobs:
...
 
+  clippy:
+    name: Cargo clippy
+    # TODO: replace `parity-default` with `ubuntu-latest` when the repo becomes public.
+    runs-on: parity-default
+    needs: [set-image]
+    container:
+      image: ${{ needs.set-image.outputs.CI_IMAGE }}
+    env:
+      SKIP_PALLET_REVIVE_FIXTURES: 1
```
bkontur, line 89: no need to, I put it this in the ./cargo/config.toml
```suggestion
```

.github/workflows/check.yml
```diff
@@ -78,6 +78,32 @@ jobs:
         run: |
           cargo check --workspace --features=runtime-benchmarks
 
+  clippy:
+    name: Cargo clippy
```
bkontur, line 82: @karolk91 it would cool, if you could add also fmt :) (ok, as follow-up)

PR #85, opened 2025-11-01
Example for using DAG-PB for storing multiple chunks

bkontur:
This PR contains example for multipart / chunked content / big files.

The code stores one file, splits into chunks and then uploads those chunks to the Bulletin.
It collects all the partial CIDs for each chunk and saves them as a custom metadata JSON file in the Bulletin.

Now we have two examples:
1. **Manual reconstruction** — return the metadata and chunk CIDs, then reconstruct the original file manually.
2. **IPFS DAG feature** —
    * converts the metadata into a DAG-PB descriptor,
    * stores the DAG descriptor proof and verify rootCID
    * stores it directly in IPFS,
    * and allows fetching the entire file using a single root CID from an IPFS HTTP gateway (for example: `http://localhost:8080/ipfs/QmW2WQi7j6c7UgJTarActp7tDNikE4B2qXtFCfLPdsgaTQ`).

### DIFF
--- a/examples/store_chunked_data.js
+++ b/examples/store_chunked_data.js
@@ -173,7 +173,7 @@ export async function buildUnixFSDag(metadataJson) {
     const blockSizes = chunks.map(c => BigInt(c.length))
     const fileData = new UnixFS({ type: 'file', blockSizes })
 
-    console.log(`🧩 Building UnixFS DAG:
+    console.log(`\n🧩 Building UnixFS DAG:
   • totalChunks: ${chunks.length}
   • blockSizes: ${blockSizes.join(', ')}`)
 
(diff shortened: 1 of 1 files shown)


examples/store_dag.js
```diff
@@ -0,0 +1,165 @@
...
+    const dagHash = await sha256.sha256.digest(dagNodeAsBytes);
+    const expectedRootCid = CID.createV1(dagPB.code, dagHash);
+    console.log(`[DAG] ExpectedRootCid: ${expectedRootCid}`);
+
+    // Store DAG-PB file with rootCID as and on-chain proof, can be custom pallet/state
+    console.log('\n============================\n5. Storing DAG-PB proof...');
+    await storeProof(api, who_pair, expectedRootCid);
+
+    // Store DAG-PB node in IPFS
+    const dagCid = await ipfs.dag.put(dagNode, {
```
dmitry-markin: Ideally, this should be stored in the transaction storage as well.

bkontur: > Ideally, this should be stored in the transaction storage as well.

Yes, right :). Originally, I was storing it with `storeProof -> store(dagFile)`, but something has not been working as I expected, but I am still experimenting with this

examples/store_dag.js
```diff
@@ -0,0 +1,165 @@
...
+        "Hello, Bulletin remote2 - " + new Date().toString(),
+        "Hello, Bulletin remote3 - " + new Date().toString()
+    ];
+    const chunks = [];
+    for (let i = 0; i < contents.length; i++) {
+        const cid = await store(api, who_pair, contents[i], nonce.addn(i));
+        console.log(`Stored data with CID${i + 1}:`, cid);
+        chunks.push({ cid, len: contents[i].length });
+    }
+    await new Promise(resolve => setTimeout(resolve, 5000));
```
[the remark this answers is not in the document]
x3c41a: Also, do I understand correctly that your example refers to [polkadot-bulletin-chain/pallets/transaction-storage](https://github.com/paritytech/polkadot-bulletin-chain?tab=readme-ov-file#polkadot-bulletin-chainpalletstransaction-storage) pallet?

examples/store_dag.js
bkontur: I finished the example with chunking - manually and/or DAG support how does that work with storing also proof for DAG: https://github.com/paritytech/polkadot-bulletin-chain/pull/85/commits/193b9282e6c254d21208b1debcb1c6fff0894e57

bkontur:
> @bkontur What do you think about extending the transaction storage with capability of storing the codec, so we can do it completely in Bulletin without the need to deploy IPFS nodes? Not sure if you are aware, but IPFS doesn't store anything in a distributed way and only serves the local data.

> we can do it completely in Bulletin without the need to deploy IPFS nodes

Just for completness, do you mean "IPFS nodes" => "IPFS DAG nodes"?

https://github.com/paritytech/polkadot-sdk/issues/10301

PR #98, opened 2025-11-13
[Blocked] Smoldot write example

x3c41a:
Blocked by https://github.com/smol-dot/smoldot/issues/2182

I just wanted to show how to construct a proper JSON-RPC call with papi abd what to put in both `method` and `params` fields.  I can add more later on if need...

### DIFF
--- a/examples/smoldot_auth_and_store.js
+++ b/examples/smoldot_auth_and_store.js
@@ -40,4 +40,4 @@ await client
         const response = await chain.nextJsonRpcResponse();
         console.log("✅ JSON-RPC response:", response);
       })
-
+      .then(() => client.terminate())


examples/smoldot_auth_and_store.js
```diff
@@ -0,0 +1,43 @@
...
+const client = smoldot.start();
+
+// Bulletin address
+const provider = new WsProvider('ws://localhost:10000');
+const api = await ApiPromise.create({ provider });
+await api.isReady;
+
+const keyring = new Keyring({ type: 'sr25519' });
+const sudo_pair = keyring.addFromUri('//Alice');
+const who_pair = keyring.addFromUri('//Alice');
```
x3c41a: maybe it's worth using Bob here

Issue #93, opened 2025-11-12
[Investigate/Research] "IPFS-lite" "light client" solutions (a.k.a. no Kubo needed)


bkontur:
Related to:
https://github.com/paritytech/polkadot-bulletin-chain/issues/33
https://github.com/paritytech/polkadot-bulletin-chain/issues/90

Investigate how we could support an IPFS HTTP gateway for dApps without relying on external IPFS gateways.

The main idea is that Bulletin nodes already support IPFS over Bitswap (p2p). With some form of IPFS DHT auto-discovery (see https://github.com/paritytech/polkadot-bulletin-chain/issues/33 and an "intermediate" HTTP/p2p layer, we could implement everything directly within the dApps — using only a light client for Bulletin (or another chain) — and avoid connecting to external IPFS gateways.

_Disclaimer: (Hope, I understood this idea well, based on discussion with @BigTava, which he had with Basti)_

bkontur:
> Is this still ongoing?

@BigTava I left a comment in the issue you linked, basically @dmitry-markin is working on smoldot part (Web3 aligned), so we won't need helia or IPFS gateways (but still we could support both and can be also use - as Web2 and not Web3 aligned).

Issue #86, opened 2025-11-04
`pallet-transaction-storage` merge to Polkadot-SDK's one


bkontur:
Relates to: https://github.com/paritytech/polkadot-bulletin-chain/issues/74 (when we go for Bulletin Westend parachain)

At the beginning of **Bulletin**, the `pallet-transaction-storage` was copied from the **Polkadot-SDK** and adapted to Bulletin’s needs (no deposits, no tokens) to handle authorizations for storage.

At some point, when we plan to deploy the **Bulletin parachain** on Polkadot, we should merge and move this functionality back into the **Polkadot-SDK**. This will simplify maintenance and release processes for Fellows and other contributors.

The original pallet supports deposits and fees using the native token for storage. The copied version supports authorizations without requiring any deposits or fees. I propose to merge and refactor both implementations so that we can cover both use cases through the pallet’s configuration. This would allow the runtime to, for example, enable authorization-based access for certain origins while requiring paid deposits for others.

## TODO

- [ ] support authorizations by Polkadot-SDK's one
- [ ] improve configurations

PR #100, opened 2025-11-14
Full IPFS CID codec/hash support - passing `CidConfig` over `TransactionExtension`

bkontur:
Implements: https://github.com/paritytech/polkadot-sdk/issues/10301
Depends on: https://github.com/paritytech/litep2p/pull/482

## TL'DR
Basically, with this feature, we won’t need any external IPFS gateway + Bulletin can handle/support multi-chunk big files as IPFS DAG-PB.
(Just a light-client IPFS p2p/http layer connected to the Bulletin node cluster: https://github.com/paritytech/polkadot-bulletin-chain/issues/93)

## Context 

Actually, `pallet-transaction-storage` was only able to index content/transactions using a Blake2b256 hash. With this PR, we can now provide a CID codec and hashing, which allows the pallet to index content using `Sha2_256` or `Keccak256` as well. This makes it IPFS-compatible—for example, it can handle and store DAG-PB content directly **without needing an additional call to the IPFS gateway**, using:

```
const dagCid = await ipfs.block.put(dagBytes, {
    format: 'dag-pb',
    mhtype: 'sha2-256'
})

or 

const dagCid = await ipfs.dag.put(
    dagNode,
    {
        storeCodec: 'dag-pb',
        hashAlg: 'sha2-256',
        pin: true
    }
);
```

## TODO
- [x] change `store_chunked_data.js` not using `ipfs.dag.put`
- [ ] check_unsigned uses:
```
Call::<T>::store { data } => Self::check_store_renew_unsigned(
					data.len(),
					|| sp_io::hashing::blake2_256(data),
					context,
				),
```

### DIFF
--- a/runtimes/bulletin-polkadot/tests/tests.rs
+++ b/runtimes/bulletin-polkadot/tests/tests.rs
@@ -460,6 +460,95 @@ fn provide_cid_codec_extension_works() {
 }
 
 #[test]
+fn preimage_authorized_storage_transactions_work() {
+	run_test(|| {
+		advance_block();
+
+		// Use relayer_signer since only relayers can submit transactions in bulletin-polkadot
+		let account = relayer_signer();
+		let data = vec![0u8; 24];
+		let content_hash = sp_io::hashing::blake2_256(&data);
+		let call =
+			RuntimeCall::TransactionStorage(TxStorageCall::<Runtime>::store { data: data.clone() });
+
+		// Not authorized (no account or preimage auth) should fail to store.
+		assert_eq!(
+			construct_and_apply_extrinsic(account.pair(), call.clone()),
+			Err(TransactionValidityError::Invalid(InvalidTransaction::Payment))
+		);
+
+		// Authorize preimage (not account).
+		assert_ok!(runtime::TransactionStorage::authorize_preimage(
+			RuntimeOrigin::root(),
+			content_hash,
+			data.len() as u64,
+		));
+
+		// Now should work via preimage authorization.
+		assert_ok_ok(construct_and_apply_extrinsic(account.pair(), call));
+
+		// Verify preimage authorization was consumed.
+		assert_eq!(
+			runtime::TransactionStorage::preimage_authorization_extent(content_hash),
+			AuthorizationExtent { transactions: 0, bytes: 0 },
+		);
+	});
+}
+
+#[test]
+fn signed_store_prefers_preimage_authorization_over_account() {
+	run_test(|| {
+		advance_block();
+
+		// Use relayer_signer since only relayers can submit transactions in bulletin-polkadot
+		let account = relayer_signer();
+		let who: AccountId = account.to_account_id();
+		let data = vec![0u8; 100];
+		let content_hash = sp_io::hashing::blake2_256(&data);
+
+		// Setup: authorize both account and preimage
+		assert_ok!(runtime::TransactionStorage::authorize_account(
+			RuntimeOrigin::root(),
+			who.clone(),
+			5,
+			500,
+		));
+		assert_ok!(runtime::TransactionStorage::authorize_preimage(
+			RuntimeOrigin::root(),
+			content_hash,
+			data.len() as u64,
+		));
+
+		// Verify both authorizations exist
+		assert_eq!(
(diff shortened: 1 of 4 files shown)


pallets/transaction-storage/src/cids.rs
```diff
@@ -0,0 +1,244 @@
...
+///
+/// If no configuration is provided, defaults are:
+/// - Codec: raw (0x55)
+/// - Hashing: Blake2b-256
+///
+/// # Errors
+/// Returns `Err(())` if multihash wrapping fails.
+pub fn calculate_cid(data: &[u8], config: Option<CidConfig>) -> Result<CidData, ()> {
+	// Determine hashing algorithm and codec
+	let (hashing, codec) = if let Some(config) = config {
```
lexnv: nit: config.map(|config| (config.hashing, config.codec).unwrap_or((Blake, 0x55))

bkontur:
> We should probably also bring the transactions-storage pallet in substrate up to date?

@dmitry-markin yes, exactly, this is what we are trying to achieve here: https://github.com/paritytech/polkadot-bulletin-chain/issues/86

raymondkfcheung (approved):

BigTava:
Thanks for the update, @bkontur. Since we wanted to move forward and integrate the remaining pieces with Nova, I went ahead and rebased the work onto the `tiago-rebase-bko-cid-format` branch. I did not review the code, but it is working well and is now live at `wss://bulletin.dotspark.app`.

bkontur:
> but it is still not working. The IPFS gateway is unable to resolve the sha-256 CID.

@BigTava hmm, let me check, did you upgrade just runtime and/or rebuild the node?

BigTava:
> but it is still not working. The IPFS gateway is unable to resolve the sha-256 CID.

Update: It is now working correctly. The issue was that I was using a SHA-256 CIDv0 instead of CIDv1. Additionally, I uploaded the .car file and attempted to retrieve it without parsing it. Since Bulletin stores raw bytes rather than files, the gateway wasn't infering the directory structure and failed to resolve the content.

examples/native_ipfs_dag_pb_chunked_data.js
```diff
@@ -0,0 +1,192 @@
...
+    assert.deepStrictEqual(expectedRootCid, rootCid);
+
+    // Read by rootCID directly over IPFS gateway, which handles download all the chunks.
+    // (Other words Bulletin is compatible)
+    console.log('🧱 DAG stored on Bulletin with CID:', rootCid.toString())
+    console.log('\n🌐 Try opening in browser:')
+    console.log(`   http://127.0.0.1:8080/ipfs/${rootCid.toString()}`)
+    console.log('   (You’ll see binary content since this is an image)')
+    console.log('')
+    console.log(`   http://127.0.0.1:8080/ipfs/${convertCid(rootCid, 0x55)}`)
```
franciscoaguirre: Same with this value

runtimes/bulletin-polkadot/tests/tests.rs
```diff
@@ -386,6 +398,68 @@ fn transaction_storage_runtime_sizes() {
...
+		));
+		assert!(!CidConfigForStore::<Runtime>::exists());
+
+		// 3. Store data WITH a custom cid_config (Sha2_256 + 0x70 codec).
+		assert_ok_ok(construct_and_apply_extrinsic_with_codec(
+			account.pair(),
+			RuntimeCall::TransactionStorage(TxStorageCall::<Runtime>::store { data: data.clone() }),
+			Some(CidConfig { codec: 0x70, hashing: HashingAlgorithm::Sha2_256 }),
+		));
+		assert!(!CidConfigForStore::<Runtime>::exists());
```
franciscoaguirre: Why does this never exist? Should it not exist before calling `on_finalize`?

runtimes/bulletin-westend/tests/tests.rs
```diff
@@ -122,6 +146,11 @@ fn construct_and_apply_extrinsic(
 	bulletin_westend_runtime::Executive::apply_extrinsic(xt)
 }
 
+fn assert_ok_ok(apply_result: ApplyExtrinsicResult) {
```
franciscoaguirre: 😂

PR #112, opened 2025-11-21
Apply Taplo to format Cargo TOMLs

raymondkfcheung:
Uses the project config to format Cargo TOML files


x3c41a (approved):

raymondkfcheung:
> How did you do that? We have zepter/taplo configs `.config/*`.

Yes, you're right. I've re-applied to use the project config.

PR #118, opened 2025-11-24
Initial version of authorize and store with PAPI, Smoldot and IPFS

x3c41a:
Fetching CID from IPFS fails for an unknown reason, output:
```
Trying to get cid:  CID(bafk2bzacecycbokdswtfpolfvwyre2lld5pxdbqvcxckq7ant2gujpbfzmsam)
Block not found directly, trying cat... context deadline exceeded
Trying to chunk cid:  CID(bafk2bzacecycbokdswtfpolfvwyre2lld5pxdbqvcxckq7ant2gujpbfzmsam)
```
even though both `authorise` and `store` transactions were `inBlock` and I waited for an extra 12 (+1) seconds for change to propagate (just in case). However, that CID is not present in IPFS.

I back-tested this logic with what we have in `authorize_and_store_papi.js`: I stored a file with that script, got CID and tried to fetch it from the new script -- everything works fine. However, when I try to fetch CID stored with Smoldot + PAPI -- it fails. 

The only thing that I know right now is that something is wrong along this path: script(PAPI tx) -> Smoldot -> Bulletin -> IPFS.
If tx is `inblock` (and I can see that block with PJS UI) it means that there is an issue between Bulletin <--> IPFS? @bkontur 

To reproduce:
- just run [these commands](https://github.com/paritytech/polkadot-bulletin-chain/tree/main/examples#run-bulletin-solochain-with---ipfs-server)
- and this new script (`node authorize_and_store_papi_smoldot.js`)
- or use `just` commands from #111

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -82,7 +82,7 @@ docker exec -it ipfs-node ipfs swarm connect /ip4/172.17.0.1/tcp/12347/ws/p2p/12
 
 ```shell
 # Runs a script that reconnects every 2 seconds
-# Defaults to Docker; use 'local' to use the local Kubo instance
+# Defaults to 'local' (local Kubo); use 'docker' for the Docker setup
 ./scripts/ipfs-reconnect-solo.sh
 ```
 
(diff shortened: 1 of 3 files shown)


examples/authorize_and_store_papi_smoldot.js
```diff
@@ -0,0 +1,178 @@
...
+            
+            const storeBlockHash = await submitAndWatch(chain, signedStoreTx, 'store');
+            console.log('✅ Stored data with CID:', cid);
+            console.log('   In block:', storeBlockHash);
+            
+            return { chain, cid };
+        })
+        .then(async ({ chain, cid }) => {
+            console.log('\n⏳ Waiting for IPFS to propagate (13 seconds)...');
+            await new Promise(resolve => setTimeout(resolve, 13000));
```
x3c41a: with the previous script I did not do it but added here just in case

PR #128, opened 2025-12-03
Align Types between Bulletin and SDK

raymondkfcheung:
Addresses #86

### DIFF
--- a/pallets/common/src/lib.rs
+++ b/pallets/common/src/lib.rs
@@ -67,7 +67,7 @@ where
 impl<AccountId, HoldReason: Encode + Decode + TypeInfo + 'static> Inspect<AccountId>
 	for NoCurrency<AccountId, HoldReason>
 {
-	type Balance = u64;
+	type Balance = u128;
 
 	fn total_issuance() -> Self::Balance {
 		Zero::zero()


pallets/transaction-storage/src/lib.rs
```diff
@@ -619,6 +619,11 @@ pub mod pallet {
 		OptionQuery,
 	>;
 
+	/// Storage period for data in blocks. Should match `sp_storage_proof::DEFAULT_STORAGE_PERIOD`
+	/// for block authoring.
+	#[pallet::storage]
```
raymondkfcheung: @bkchr Shall we use `#[pallet::storage]` like on the SDK repo or keep as `#[pallet::constant]`?

pallets/transaction-storage/src/lib.rs
```diff
@@ -76,6 +86,128 @@ pub struct AuthorizationExtent {
 	pub bytes: u64,
 }
 
+/// A no-op fungible implementation for this pallet.
+///
+/// All balances are always zero and every mutation is accepted but ignored. This is useful for
+/// runtimes that do not want to charge any currency like Bulletin.
+pub struct NoopCurrency<Reason>(PhantomData<Reason>);
```
bkontur: can we just reuse `pallets-common` impl here?

raymondkfcheung: Yes, updated: https://github.com/paritytech/polkadot-bulletin-chain/pull/128/commits/2004df19699ce88ea082e5ad2037199fdf83d019

pallets/transaction-storage/src/lib.rs
raymondkfcheung: For `StoragePeriod`, I’ll follow up on this in a separate PR.

PR #141, opened 2025-12-12
Chunked data example + nits

bkontur:
Relates to: https://github.com/paritytech/polkadot-bulletin-chain/issues/142

### DIFF
--- a/examples/common.js
+++ b/examples/common.js
@@ -114,6 +114,9 @@ export async function buildUnixFSDagPB(chunks, mhCode = 0x12) {
     return { rootCid, dagBytes }
 }
 
+/**
+ * Generates (dynamic) images based on the input text.
+ */
 export function generateTextImage(file, text, width = 800, height = 600) {
     const canvas = createCanvas(width, height);
     const ctx = canvas.getContext("2d");


examples/store_chunked_data.js
```diff
@@ -287,84 +248,99 @@ async function authorizeStorage(api, sudoPair, pair, nonceMgr) {
...
+        console.log('🛰 Connecting to Bulletin node...')
+        const provider = new WsProvider(WS_ENDPOINT)
+        api = await ApiPromise.create({ provider })
+        await api.isReady
+        const ipfs = create({ url: IPFS_API });
+        console.log('✅ Connected to Bulletin node')
+
+        const keyring = new Keyring({ type: 'sr25519' })
+        const pair = keyring.addFromUri('//Alice')
+        const sudoPair = keyring.addFromUri('//Alice')
```
x3c41a: we repeat this pattern across all our test scripts. However, it's only 8 lines of code. The refactoring might not pay off. 
I just want to confirm that we keep replicating this patter across future scripts too?

examples/store_chunked_data.js
```diff
@@ -287,84 +248,99 @@ async function authorizeStorage(api, sudoPair, pair, nonceMgr) {
...
+        ////////////////////////////////////////////////////////////////////////////////////
+        // 1. example manually retrieve the picture (no IPFS DAG feature)
+        const metadataJson = await retrieveMetadata(ipfs, metadataCid)
+        await retrieveFileForMetadata(ipfs, metadataJson, OUT_1_PATH);
+        filesAreEqual(FILE_PATH, OUT_1_PATH);
+
+        ////////////////////////////////////////////////////////////////////////////////////
+        // 2. example download picture by rootCID with IPFS DAG feature and HTTP gateway.
+        // Demonstrates how to download chunked content by one root CID.
+        // Basically, just take the `metadataJson` with already stored chunks and convert it to the DAG-PB format.
```
x3c41a: 1. Do we want to showcase that one solution is faster than the other (or any other relevant metrics)?

2. By looking at the example itself, I don't really understand when I should use example 1 and when example 2. Maybe it's worth explaining (or linking to the docs) what `IPFS DAG feature and HTTP gateway` give us?

3. dq: can we also see `http://127.0.0.1:8080/ipfs/${rootCid.toString()}` for the first example?

examples/common.js
```diff
@@ -64,3 +69,104 @@ export function setupKeyringAndSigners(sudoSeed, accountSeed) {
...
+export function generateTextImage(file, text, width = 800, height = 600) {
+    const canvas = createCanvas(width, height);
+    const ctx = canvas.getContext("2d");
+
+    // 🎨 Background
+    ctx.fillStyle = randomColor();
+    ctx.fillRect(0, 0, width, height);
+
+    // 🟠 Random shapes
+    for (let i = 0; i < 15; i++) {
```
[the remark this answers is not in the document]
bkontur: this function (vibe-coded :)) is really no important, I just needed dynamic images to eliminate accidentally IPFS cached old content,

examples/common.js
```diff
@@ -64,3 +69,104 @@ export function setupKeyringAndSigners(sudoSeed, accountSeed) {
...
+
+    // Hash DAG to produce CIDv1
+    const rootCid = await cidFromBytes(dagBytes, dagPB.code, mhCode)
+
+    console.log(`✅ DAG root CID: ${rootCid.toString()}`)
+
+    return { rootCid, dagBytes }
+}
+
+export function generateTextImage(file, text, width = 800, height = 600) {
```
bkontur:
```suggestion
/**
 * Generates (dynamic) images based on the input text.
 */
export function generateTextImage(file, text, width = 800, height = 600) {
```

examples/common.js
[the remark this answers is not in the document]
bkontur: I added docs at least: https://github.com/paritytech/polkadot-bulletin-chain/pull/141/commits/fbd0813092a4468fe2eed32d1267d01298cdabc2

PR #147, opened 2025-12-15
setup github matrix feature

x3c41a:
Also parametrised functions in `justfile` to support different runtimes. 

## Proposal
If it comes to [runtimes-matrix](https://github.com/paritytech/polkadot-bulletin-chain/blob/main/scripts/runtimes-matrix.json#L24) values and their corresponding `.toml` configs -- they're inconsistent and it creates a mess, especially for [rococo](https://github.com/paritytech/polkadot-bulletin-chain/blob/main/scripts/runtimes-matrix.json#L24) `name` field is inconsistent with `package` field.

The values of `runtimes-matrix.json` are used in both [cmd.py](https://github.com/paritytech/polkadot-bulletin-chain/blob/main/scripts/cmd/cmd.py#L152) and `.github/workflows` now.
To avoid maintenance burden I'd suggest:
- [1] either we introduce some naming convention to `name` field so that I can re-use it to prepend `.toml` (for instance),
- [2] or we introduce a new `toml_path` field (exact name TBD) in order not to maintain one in the `justfile`

WDYT?

### DIFF
--- a/examples/justfile
+++ b/examples/justfile
@@ -41,7 +41,7 @@ setup-westend-prerequisites:
     
     # Build polkadot binary
     echo "   Building polkadot binary (this may take a while)..."
-    cargo build -p polkadot -r
+    DYLD_FALLBACK_LIBRARY_PATH=/Library/Developer/CommandLineTools/usr/lib cargo build -p polkadot -r
     
     # Verify and copy polkadot binaries
     echo "   Copying polkadot binaries..."
@@ -55,7 +55,7 @@ setup-westend-prerequisites:
     
     # Build polkadot-parachain binary
     echo "   Building polkadot-parachain binary (this may take a while)..."
-    cargo build -p polkadot-parachain-bin -r
+    DYLD_FALLBACK_LIBRARY_PATH=/Library/Developer/CommandLineTools/usr/lib cargo build -p polkadot-parachain-bin -r
     
     # Verify and copy polkadot-parachain binary
     echo "   Copying polkadot-parachain binary..."
@@ -67,7 +67,7 @@ setup-westend-prerequisites:
     
     # Build and install chain-spec-builder
     echo "   Building chain-spec-builder..."
-    cargo build -p staging-chain-spec-builder -r
+    DYLD_FALLBACK_LIBRARY_PATH=/Library/Developer/CommandLineTools/usr/lib cargo build -p staging-chain-spec-builder -r
     
     # Verify and copy chain-spec-builder binary
     echo "   Copying chain-spec-builder binary..."
@@ -164,7 +164,7 @@ bulletin-solo-zombienet-start runtime="bulletin-polkadot-runtime":
     else # bulletin-polkadot-runtime
         echo "   Setting up Polkadot runtime..."
         # Build the node binary (which will build the runtime as a dependency)
-        cargo build --release -p polkadot-bulletin-chain
+        DYLD_FALLBACK_LIBRARY_PATH=/Library/Developer/CommandLineTools/usr/lib cargo build --release -p polkadot-bulletin-chain
         BULLETIN_BINARY=$(cd .. && pwd)/target/release/polkadot-bulletin-chain
         ZOMBIENET_CONFIG=$(cd .. && pwd)/zombienet/bulletin-polkadot-local.toml
         cd ..


examples/justfile
```diff
@@ -6,11 +6,82 @@
...
-build:
-    cargo build --release -p polkadot-bulletin-chain
+# Setup prerequisites for Westend runtime (polkadot and polkadot-parachain binaries)
+# This recipe clones polkadot-sdk, builds required binaries, and copies them to ~/local_bridge_testing/bin
+setup-westend-prerequisites:
+    #!/usr/bin/env bash
+    set -e
+    
+    BIN_DIR=~/local_bridge_testing/bin
+    POLKADOT_SDK_DIR=~/local_bridge_testing/polkadot-sdk
```
bkontur:
```suggestion
    BIN_DIR=~/local_bulletin_testing/bin
    POLKADOT_SDK_DIR=~/local_bulletin_testing/polkadot-sdk
```

examples/justfile
```diff
@@ -6,11 +6,82 @@
...
+    
+    echo "🔧 Setting up Westend runtime prerequisites..."
```
bkontur:
```suggestion
    echo "🔧 Setting up parachain runtime prerequisites..."
```

examples/justfile
```diff
@@ -6,11 +6,82 @@
...
+    ls -la target/release/chain-spec-builder
+    cp target/release/chain-spec-builder $BIN_DIR/
+    
+    # Verify chain-spec-builder
+    $BIN_DIR/chain-spec-builder --version
+    
+    # Add BIN_DIR to PATH for subsequent scripts
+    export PATH="$BIN_DIR:$PATH"
+    
+    echo "✅ Westend prerequisites setup complete!"
```
bkontur:
```suggestion
    echo "✅ Parachain prerequisites setup complete!"
```

examples/justfile
```diff
@@ -6,11 +6,82 @@
...
+    
+    # Check out the required branch
+    echo "   Checking out branch bko-bulletin-para-support..."
+    git fetch origin
+    # TODO: later, before merging, switch to the `bko-bulletin-support` https://github.com/paritytech/polkadot-sdk/pull/10662 (not now)
+    git reset --hard origin/bko-bulletin-para-support
+    
+    # Build polkadot binary
+    echo "   Building polkadot binary (this may take a while)..."
+    DYLD_FALLBACK_LIBRARY_PATH=/Library/Developer/CommandLineTools/usr/lib cargo build -p polkadot -r
```
bkontur: if `DYLD_FALLBACK_LIBRARY_PATH=/Library/Developer/CommandLineTools/usr/lib` is really needed (ok, for macos local), let's do this just once at the beginning like:

```
if [[ "$OSTYPE" == "darwin"* ]]; then
      export DYLD_FALLBACK_LIBRARY_PATH=/Library/Developer/CommandLineTools/usr/lib
fi
```

examples/justfile
```diff
@@ -28,14 +99,16 @@ _ipfs-cmd:
         exit 1
     fi
 
-# Start start solo chain with zombienet in background
-bulletin-solo-zombienet-start:
+# Start solo chain or parachain with zombienet in background
+# Parameters:
+#   runtime - Runtime name (e.g., "bulletin-polkadot-runtime", "bulletin-westend-runtime"), taken from ../scripts/runtimes-matrix.json
+bulletin-solo-zombienet-start runtime="bulletin-polkadot-runtime":
```
bkontur: I don't know, I would avoid setting here defauls, anyway we will switch to the `bulletin-westend-runtime"` probably, but let's keep this clean without default for now
```suggestion
bulletin-solo-zombienet-start runtime="":
```

examples/justfile
```diff
@@ -167,18 +297,20 @@ papi-generate:
...
     echo "🔧 Tearing down Docker services if they are running..."
     just ipfs-shutdown
     
-    echo "🐳 Setting up services with Docker IPFS..."
+    echo "🐳 Setting up services with Docker IPFS (runtime: {{ runtime }})..."
     
     # Start services with Docker
     just ipfs-start
-    just bulletin-solo-zombienet-start
+    just bulletin-solo-zombienet-start {{ runtime }}
```
bkontur: And `network` or `network-setup` could be "solo" or "para" value specified in runtimes-matrix.json

```suggestion
    if [ "{{ network }}" = "solo" ]; then
        just bulletin-solo-zombienet-start {{ runtime }}
    if [ "{{ network }}" = "para" ]; then
        just bulletin-para-zombienet-start {{ runtime }}    
    fi        
```

examples/justfile
x3c41a: I have a mixed feeling about this change, tbh. 

We still explicitly provide all the parameters when we call a `just` function, and defaults is one of `just` features. 
I don't have a strong opinion here. This change won't change much, sounds like a "nit", unless you have some strong arguments to back it up.

> but let's keep this clean without default for now

also there is a `default` function -- `just default`. Any preferences there?

x3c41a:
I am really concerned about the size and complexity of this PR (not because it's hard but because it's complex!); we already had this problem before. 
To maintain code quality and make reviews useful, I'd strongly prefer we move toward smaller, incremental updates rather than large, multi-goal changes!!!

This PR was supposed to be enabling matrix feature only, but now it's:
- enabling matrix feature
- setting up Westend pipeline on CI
  - fixing Westend pipeline on CI because it does not work,
  - fixing Westend pipeline on Mac because it does not work too,
 - parametrising and refactoring `just`

With the amount of changes like this in a single PR it's hard to make a focused effort towards one goal.
Instead of making small incremental changes, having them tested, approved and merged, I’m finding myself losing the original context while troubleshooting new issues, which in turn may create other issues and so on and so forth

I put this PR on hold and start the split

examples/justfile
x3c41a: > also the zombienet file ( /zombienet/bulletin-westend-local.toml or solo) should be specified in runtimes-matrix.json

I am not sure if you mean File name or Path to a File? I think Path needs to be relative because working directory may differ. I'd leave as is (`ZOMBIENET_CONFIG=$(cd .. && pwd)/zombienet/bulletin-westend-local.toml`)

> and also chain spec command could be as a part of runtimes-matrix.json (./scripts/create_bulletin_westend_spec.sh)

Working directory is `examples` so path needs to be `.././scripts/create_bulletin_westend_spec.sh`.

What I would do, I would decouple `PATH` from `filename` and handle them separately:
- PATH stays relative as is in the `justfile`
- add two new properties to `runtimes_matrix`: `zombienet_file_name` and `chain_spec_script_name`
This way we can have both relative path and configurable file names

WDYT?

x3c41a:
Obsolete

PR #150, opened 2025-12-17
chore: replace deprecated `ws_port` with `rpc_port`

peetzweg:
In our pop testnet setup we still use this config and i sometimes stumble over this again sometimes.


bkontur:
@peetzweg continuation here: https://github.com/paritytech/devops/issues/4686

PR #170, opened 2026-01-12
Fixed smoldot westend integration test

x3c41a:
# Key Fixes

## Smoldot Bootnode Issue
  - Problem: Smoldot light client couldn't connect to parachain nodes. Smoldot supports only WebSocket transport, but zombienet bootnodes used only TCP.
  - Fix: I added `convertBootNodeToWebSocket()` that converts TCP `multiaddrs` to WebSocket format using **convention** `ws_port = tcp_port + 1`. Zombienet nodes configured with `--listen-addr /ip4/0.0.0.0/tcp/{port+1}/ws`

## Transaction Indexing
  - Problem: Parachain wasn't properly indexing transactions for the `TransactionStorage` pallet,
  - Fix: Enabled transaction-index feature on cumulus-pallet-parachain-system:
`  cumulus-pallet-parachain-system = { features = ["transaction-index"], workspace = true }`
@bkontur 

## CI apt-get 403 Errors
  - Problem: Microsoft package repos on Github runners returned `403 Forbidden`
  - Fix: Removed Microsoft repo lists before apt-get update

 ## Other Changes
  - Added waitForChainReady to verify chain accessibility
  - Refactored smoldot script for parachain support (relay + para chain specs)

### DIFF
--- a/scripts/create_bulletin_westend_spec.sh
+++ b/scripts/create_bulletin_westend_spec.sh
@@ -14,5 +14,4 @@ chain-spec-builder create \
         -r ./target/release/wbuild/bulletin-westend-runtime/bulletin_westend_runtime.compact.compressed.wasm \
         named-preset local_testnet
 
-mv chain_spec.json bulletin-westend-spec.json
-cp bulletin-westend-spec.json ./zombienet/bulletin-westend-spec.json
+mv chain_spec.json ./zombienet/bulletin-westend-spec.json


examples/justfile
```diff
@@ -387,11 +387,14 @@ run-authorize-and-store runtime mode="ws": npm-install
     if [ "{{ mode }}" = "smoldot" ]; then
         # Set chainspec path based on runtime
         if [ "{{ runtime }}" = "bulletin-westend-runtime" ]; then
-            CHAINSPEC_PATH="$TEST_DIR/bulletin-westend-collator-1/cfg/westend-local-1006.json"
-        else # bulletin-polkadot-runtime
+            # Parachain: relay chain (required) + parachain spec (optional)
+            RELAY_CHAINSPEC_PATH="$TEST_DIR/bob/cfg/westend-local.json"
+            PARACHAIN_CHAINSPEC_PATH="$TEST_DIR/bulletin-westend-collator-1/cfg/westend-local-1006.json"
```
bkontur: @x3c41a I think this could be a problem, using collator-1, because collator-1 does not have bootstrap peers, this is why we use bob instead of alice for relay/solo chain:
https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/20916008493/job/60089955004?pr=170#step:12:130
```
[smoldot:WARN] smoldot: Newly-added chain bulletin-westend has an empty list of bootnodes. Smoldot will likely fail to connect to its peer-to-peer network.
```

```
```

```suggestion
            PARACHAIN_CHAINSPEC_PATH="$TEST_DIR/bulletin-westend-collator-2/cfg/westend-local-1006.json"
```

zombienet/bulletin-westend-local.toml
```diff
@@ -15,14 +15,20 @@ chain = "westend-local"
 [[relaychain.nodes]]
 name = "alice"
 validator = true
+p2p_port = 30333
 rpc_port = 9942
 balance = 2000000000000
+# WebSocket P2P on p2p_port + 1 for smoldot light client support
+args = ["--listen-addr", "/ip4/0.0.0.0/tcp/30334/ws"]
```
bkontur: `p2p_port=30333` does not work? Why we need to use `+1` 30334?

.github/workflows/integration-test.yml
```diff
@@ -154,26 +156,22 @@ jobs:
           echo "ZOMBIENET_BINARY=zombienet-linux-x64" >> "$GITHUB_ENV"
 
       # Westend parachain
-      - name: Run authorize and store (PAPI, RPC node, Westend parachain)
+      - name: Run authorize and store (PAPI, smoldot, Westend parachain)
```
bkontur: @x3c41a I think you just change the order for Westend right? Please let's keep the same test/order for every env/chain

examples/api.js
```diff
@@ -37,42 +37,61 @@ export const TX_MODE_IN_BLOCK = "in-block";
...
-                            console.log(`📦 ${txName} included in finalized block:`, ev.block.hash);
-                            sub.unsubscribe();
-                            resolve(ev);
-                        }
-                        break;
-
-                    default:
-                        throw new Error("Unhandled txMode: " + txMode)
+                if (!resolved && config.match(ev)) {
+                    console.log(config.log(txName, ev));
```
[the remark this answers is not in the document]
bkontur: ahaaaa, omg, my fault, I shouldn't be reviewing late night :D
I saw `console.log(console.log ` :D

x3c41a:
> remove `bulletin-westend-spec.json` file

I also thought that it was redundant but it actually contains account balance overwrites which allow Alice (IIRC) to sign and send/spend the transaction. Otherwise `Westend DAG` test fails

x3c41a:
Shall I merge? @bkontur

Issue #174, opened 2026-01-14
CI improvement


x3c41a:
Parametrise all [examples](https://github.com/paritytech/polkadot-bulletin-chain/tree/main/examples) tests with `network_address` and private key (eventually through `.env` but can be a parameter at first).
Then propagate the changes all the way to the CI, through `just` and intermediary functions.

PR #182, opened 2026-01-16
Optimised CI

x3c41a:
Another attempt on a fresh master (no merge conflicts this time)

### DIFF
--- a/examples/justfile
+++ b/examples/justfile
@@ -234,7 +234,7 @@ ipfs-start test_dir: _check-docker
     echo "   Container: ipfs-node"
     echo "   Waiting for container to start..."
     sleep 5
-    
+
     # Bitswap logging
     docker logs -f ipfs-node > {{ test_dir }}/ipfs-node.log 2>&1 &
     docker exec ipfs-node ipfs log level bitswap debug


examples/justfile
```diff
@@ -360,10 +360,100 @@ teardown-services test_dir:
     just ipfs-shutdown "{{ test_dir }}"
     echo "✅ Docker services stopped"
 
+# File to store current test directory path for multi-test runs
+_test-dir-file := "/tmp/bulletin-test-dir"
+
+# Start all services for a runtime (use with run-test-* recipes)
+# Usage: just start-services bulletin-polkadot-runtime
+start-services runtime: npm-install
```
bkontur: I would add test_dir as param
```suggestion
start-services test_dir runtime: npm-install
```

examples/authorize_and_store_papi_smoldot.js
```diff
@@ -120,8 +121,9 @@ async function main() {
         const bulletinAPI = client.getTypedApi(bulletin);
         await waitForChainReady(bulletinAPI);
 
-        // Signers.
-        const { sudoSigner, whoSigner, whoAddress } = setupKeyringAndSigners('//Alice', '//Alice');
+        // Signers: Use Bob for the account being authorized to avoid nonce conflicts
+        // when running after ws test (which uses Alice) on the same chain.
+        const { sudoSigner, whoSigner, whoAddress } = setupKeyringAndSigners('//Alice', '//Bob');
```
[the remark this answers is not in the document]
bkontur: hmm, I would say that every test runs after new block, so the nonce shouldn't be a problem, which test fails exactly?

Here we have some kind of NonceManger to support this
https://github.com/paritytech/polkadot-bulletin-chain/blob/00aa68c61283b3eae231b5fd5fe33346c49887a9/examples/store_chunked_data.js#L54

And also for big data, I avoid this problem with multiple accounts, so switching to `Bob` is ok:
https://github.com/paritytech/polkadot-bulletin-chain/blob/00aa68c61283b3eae231b5fd5fe33346c49887a9/examples/store_big_data.js#L147


Maybe if we just change the tests, so every test is using different account like `//Account1` `//Account2` instead of useing `//Alice` everywhere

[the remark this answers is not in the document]
x3c41a: > which test fails exactly?

Tests were failing in a random order for different reasons throughout the day. It's hard to identify a single one
https://github.com/paritytech/polkadot-bulletin-chain/pull/179

> I would say that every test runs after new block 

I would say it too but ....

> Here we have some kind of NonceManger to support this

Maybe I try to implement it as a follow-up PR? 
We need a long term solution so that we don't create a new account per test.
WDYT? @bkontur

examples/justfile
x3c41a: done

examples/justfile
x3c41a: JFYI
After this PR I have a feeling that it's high time to refactor justfile - https://github.com/paritytech/polkadot-bulletin-chain/issues/185

examples/justfile
```diff
@@ -365,10 +365,100 @@ teardown-services test_dir:
...
+_test-dir-file := "/tmp/bulletin-test-dir"
+
+# Start all services for a runtime (use with run-test-* recipes)
+# Usage: just start-services /tmp/my-test-dir bulletin-polkadot-runtime
+start-services test_dir runtime: npm-install
+    #!/usr/bin/env bash
+    set -e
+    mkdir -p "{{ test_dir }}"
+    echo "{{ test_dir }}" > {{ _test-dir-file }}
+    echo "📁 Test directory: {{ test_dir }} (stored in {{ _test-dir-file }})"
```
bkontur:
```suggestion
    echo "📁 Starting services - test directory: {{ test_dir }}"
```

examples/justfile
```diff
@@ -365,10 +365,100 @@ teardown-services test_dir:
...
+    #!/usr/bin/env bash
+    if [ ! -f "{{ _test-dir-file }}" ]; then
+        echo "❌ Error: No test directory file found at {{ _test-dir-file }}"
+        echo "   Did you run 'just start-services <runtime>' first?"
+        exit 1
+    fi
+    TEST_DIR="$(cat {{ _test-dir-file }})"
+    echo "📁 Using test directory: $TEST_DIR"
+    just teardown-services "$TEST_DIR"
+    rm -f "{{ _test-dir-file }}"
```
bkontur:
```suggestion
# Stop all services)
stop-services test_dir:
    #!/usr/bin/env bash
    echo "📁 Stopping services - test directory: {{ test_dir }}"
    just teardown-services "{{ test_dir }}"
    rm -f "{{ test_dir }}"
```

examples/justfile
```diff
@@ -365,10 +365,100 @@ teardown-services test_dir:
     just ipfs-shutdown "{{ test_dir }}"
     echo "✅ Docker services stopped"
 
+# File to store current test directory path for multi-test runs
+_test-dir-file := "/tmp/bulletin-test-dir"
```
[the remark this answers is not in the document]
x3c41a: done

examples/authorize_and_store_papi.js
```diff
@@ -7,20 +7,26 @@ import { setupKeyringAndSigners } from './common.js';
 import { cidFromBytes } from "./cid_dag_metadata.js";
 import { bulletin } from './.papi/descriptors/dist/index.mjs';
 
-const NODE_WS = 'ws://localhost:10000';
+// Command line arguments: [ws_url] [seed]
+const args = process.argv.slice(2);
+const NODE_WS = args[0] || 'ws://localhost:10000';
```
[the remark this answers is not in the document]
bkontur: Yes, yes, but the BW was deployed just yesterday and I was so desperate to test it live to see.
But I updated desc #174 - which we really need now.

bkontur (approved):
@x3c41a very good job, from 42 to 24 minutes (cca 20 minutes is just building runtime or node)

image

Issue #167, opened 2026-01-09
[Example] Authorize preimage and store


bkontur:
## TODO
- [x] Create `authorize_preimage_and_store_papi.js` (copy of `authorize_and_store_papi.js`)
- [x] Instead of `TransactionStorage.authorize_account` we want `TransactionStorage.authorize_preimage`
- [x] we need to submit `store(data)` as unsigned transaction, so no signer (anybody can submit)
  - Investigate how to do it with PAPI: `export async function store(typedApi, signer, data) {` and `function waitForTransaction(tx, signer` -> `tx.signSubmitAndWatch(signer)`
- [x] add it to the `examples/justfile`
- [x] add it to the `.github/workflows/integration-test.yml`

PR #186, opened 2026-01-16
[Example] Authorize preimage and store

rosarp:
Adding example as per #167

### DIFF
--- a/examples/authorize_preimage_and_store_papi.js
+++ b/examples/authorize_preimage_and_store_papi.js
@@ -27,7 +27,7 @@ async function main() {
         let expectedCid = await cidFromBytes(dataToStore);
         let contentHash = getContentHash(dataToStore);
 
-        // Authorize an account.
+        // Authorize a preimage.
         await authorizePreimage(
             bulletinAPI,
             sudoSigner,


examples/authorize_preimage_and_store_papi.js
```diff
@@ -0,0 +1,67 @@
...
+
+        // Data to store.
+        const dataToStore = "Hello, Bulletin with PAPI - " + new Date().toString();
+        let expectedCid = await cidFromBytes(dataToStore);
+
+        // Authorize an account.
+        await authorizePreimage(
+            bulletinAPI,
+            sudoSigner,
+            BigInt(dataToStore.length)
```
bkontur: authorizePreimage takes bytes, right?
```suggestion
            dataToStore
```

examples/api.js
```diff
@@ -46,6 +47,32 @@ export async function authorizeAccount(
     }
 }
 
+export async function authorizePreimage(
+    typedApi,
+    sudoSigner,
+    bytes,
```
bkontur: @rosarp let's change `bytes` to `content_hashes` so we can authorize multiple content_hashes with this function
```suggestion
    content_hashes,
```

Similar as we have for user authorizations above - we can authorize multiple at once:
```
const accounts = Array.isArray(whos) ? whos : [whos];
```

---


Please, create new function - which will calculate content_hash from bytes:
```
export function content_hash(bytes, mhCode = 0xb220)  { swith(mhCode) case 0xb220 => blake2AsU8a(bytes) ...}
```
And reuse it here also:
https://github.com/paritytech/polkadot-bulletin-chain/blob/7df41748d64036ff024b0bba70e60c123d6f2d3a/examples/cid_dag_metadata.js#L65

examples/common.js
```diff
@@ -226,6 +227,10 @@ export async function waitForChainReady(typedApi, maxRetries = 10, retryDelayMs
     return false;
 }
 
+export function contentHash(bytes) {
```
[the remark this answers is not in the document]
rosarp: I have removed mhCode from this. so switch is not required.
this dont need multihash right?

rosarp:
> @rosarp please rebase/merge with main

rebase completed

PR #189, opened 2026-01-19
Simplify release workflow to build and upload runtime wasms

bkontur:
Replace the previous obsolete release workflow with a simpler runtime-focused workflow that:
- Triggers on v* tags
- Builds bulletin-westend and bulletin-polkadot runtimes using srtool
- Generates release notes with runtime info (size, hash, version, etc.)
- Uploads compressed wasm files to GitHub release

# TODO
- [x] try parity-large (also for testing)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,7 +10,7 @@ permissions:
 
 jobs:
   build-runtimes:
-    runs-on: parity-default
+    runs-on: ubuntu-latest
     strategy:
       matrix:
         include:


.github/workflows/check.yml
```diff
@@ -107,9 +107,9 @@ jobs:
 
   test:
     name: Test
-    runs-on: parity-default
+    runs-on: parity-large
     timeout-minutes: 60
-    needs: [set-image, check]
```
x3c41a: dq: Doesn't `check` include `check-fmt` as one of its commands?

.github/workflows/release.yml
```diff
@@ -1,154 +1,102 @@
...
 
-permissions:
-  contents: read
-
-# Controls when the action will run.
 on:
   push:
-    # Triggers the workflow on tag push events
     tags:
-      - v[0-9]+.*
```
x3c41a: Isn't `v*` kind of too generic? 
I, personally, see the value of `v[0-9]+.*` in filtering out some unexpected tags which may occur. 

I'd vote for more strict validation rules `v[0-9]+.*` but I am also eager to hear your reasoning! @bkontur

.github/workflows/release.yml
```diff
@@ -1,154 +1,102 @@
...
+
+      - name: Add runtime info to changelog
+        run: |
+          CONTEXT=$(find . -name '*_srtool_output.json')
+          SRTOOL() { <$(<<<$CONTEXT head -n1) jq -r .$1; }
+          WASM() { <${JSON} jq -r ".runtimes.compressed.subwasm.$1"; }
+
+          tee RELEASE_NOTES.md <<-EOF
+          # Runtime info
+          *These runtimes were built with **$(SRTOOL rustc)** using **[$(SRTOOL gen)](https://github.com/paritytech/srtool)***
```
x3c41a: is it needed or auto-generated?

PR #195, opened 2026-01-22
Review bulletin westend's xcm config

franciscoaguirre:
Addresses part of https://github.com/paritytech/polkadot-bulletin-chain/issues/168

- No DOT teleports, DOT reserve transfers
- Remove superuser from the relay
- Removed unnecessary barrier
- Changed Bulletin's para id to that of a non-system parachain

# TODO

- [x] test that reserve transfers of DOT between Asset Hub and Bulletin work
- [ ] fresh weights for bulletin-westend
- [x] change 1006 paraId where possible to 2487

PR #200, opened 2026-01-26
Adding bulletin westend benchmarks

rosarp:
Adds benchmarks mentioned in #195


bkontur:
@rosarp you mentioned some bug in cmd.py yesterday? Is it still relevant?

PR #237, opened 2026-02-12
fix(ci): route IPFS CLI commands through daemon API to avoid repo.lock

bkontur:
The IPFS Docker container starts the daemon automatically, which holds repo.lock. Running `ipfs config` commands directly contends for the same lock and fails intermittently on slow CI runners.

Route all CLI commands through the daemon's HTTP API via `--api` flag, and replace fixed `sleep` waits with polling loops that check API readiness.

E,g: https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/21937803360/job/63355998706?pr=202#step:22:491


bkontur:
@x3c41a maybe matrix and separate run per runtime would also fix: https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/21937803360/job/63355998706?pr=202#step:22:491

examples/justfile
```diff
@@ -240,34 +240,60 @@ ipfs-start test_dir: _check-docker
     # Start container - daemon will init and start
     docker run -d --name ipfs-node -v ipfs-data:/data/ipfs $NETWORK_ARGS ipfs/kubo:latest
     echo "   Container: ipfs-node (network: $NETWORK_ARGS)"
-    echo "   Waiting for IPFS to initialize..."
-    sleep 5
```
x3c41a: nice! no more IPFS sleeps. 

we shall remove the remaining sleeps one day too

PR #250, opened 2026-02-18
adding storage faucet for user

### LINKED ISSUE #248: [UI] Add storage faucet to the Bulletin UI
- add panel/tab to the Bulletin UI
- three inputs:
  - for_who -> user passes ss58 accountId string 
    - after input is filled, the UI automatically displays actual authorizaion for that user
  - transactions -> number of transactions (prefilled 100)
  - bytes -> number of bytes (prefilled with 10 MiB - 10 * 1024 * 1024)
- button "authorize user"
  - in backgroud this will issue sudo call authorize_user(for_who, transactions, bytes) signed with hard-coded `//Alice` in the code (we don't care about testnets)
  - loader, when finalized it should change the actual displayed user authorizations

rosarp:
Closes #248

### DIFF
--- a/console-ui/src/pages/Authorizations/Authorizations.tsx
+++ b/console-ui/src/pages/Authorizations/Authorizations.tsx
@@ -1145,7 +1145,7 @@ function StorageFaucetTab() {
             {aliceBalance !== null && aliceBalance === 0n && (
               <div className="p-3 rounded-md bg-amber-500/10 border border-amber-500/20 text-amber-600 dark:text-amber-400">
                 <AlertCircle className="h-4 w-4 inline mr-2" />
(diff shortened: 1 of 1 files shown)


bkontur (changes requested):
@rosarp please, update, I pushed some nits and cosmetics to the UI in general, but please check:

- [ ] add also panel for "Authorize Preimage"
  - [ ] required input for blake2 hash
  - [ ] text area or file upload -> when you add here anything it will just calculate blake2 hash and fills the input
  - [ ] submit button, only enabled when input with blake2 is filled up
- [ ] please, remove `sudo`, my mistake, it should be regular call for Alice, but you need to send some token to Alice account from Paseo faucet
- [ ] Add here screen, that you pass the authorization for Bob account :)

PR #275, opened 2026-02-26
feat: add /format skill for pre-commit formatting checks

mudigal:
## Summary
- Add new `/format` Claude skill that runs all formatting and cleaning tasks (Rust fmt, TOML formatting, zepter, clippy)
- Update CLAUDE.md to instruct Claude to automatically run `/format` after generating code and before commits

## Test plan
- [x] Verify `/format` skill is available in Claude Code
- [x] Test that running `/format` executes all formatting commands

### DIFF
--- a/pallets/transaction-storage/src/lib.rs
+++ b/pallets/transaction-storage/src/lib.rs
@@ -1040,13 +1040,13 @@ pub mod pallet {
 			};
 
 			if consume {
-				if Authorizations::<T>::mutate(&scope, consume_authorization)? {
-					Self::authorization_removed(&scope);
+				if Authorizations::<T>::mutate(scope, consume_authorization)? {
+					Self::authorization_removed(scope);
 				}
 			} else {
 				// Note we call consume_authorization on a temporary; the authorization in storage
 				// is untouched and doesn't actually get consumed
-				let mut authorization = Authorizations::<T>::get(&scope);
+				let mut authorization = Authorizations::<T>::get(scope);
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -6,6 +6,11 @@
 - NEVER add Co-Authored-By lines to commits
 - NEVER use git push --force or git push -f
 
+**Automatic formatting:**
+- ALWAYS run `/format` after generating or modifying Rust code
+- ALWAYS run `/format` before creating any git commit
+- This ensures all code follows project formatting standards (Rust, TOML, feature propagation) and passes clippy
```
karolk91: Wouldn't these be enough without the SKILL.md ?

x3c41a: > does separate skill bring any value?

separation of concerns and composability, the latter is the strongest value proposition, I'd say

.claude/skills/format/SKILL.md
```diff
@@ -0,0 +1,37 @@
...
+   ```
+
+## Notes
+
+- Run formatting commands (steps 1-3) first as they may auto-fix issues
+- Clippy warnings should be treated as errors (`-D warnings`)
+- If `taplo` or `zepter` are not installed, inform the user how to install them:
+  - `cargo install taplo-cli`
+  - `cargo install zepter`
+- Report all errors found and fix them where possible
```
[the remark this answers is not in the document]
x3c41a: The instruction you suggest is not future proof, it's too specific. Let's keep it generic (as is)

PR #290, opened 2026-02-27
Remove Sudo.sudo from examples - use direct signing

### LINKED ISSUE #288: [Examples] Do not use `sudo` for all authorizations in example - we have `//Alice` in genesis
## TODO

- [ ] basically remove `typedApi.tx.Sudo.sudo` for authorizations
- [ ] rename sudoSigner -> authorizationSigner

x3c41a:
## Summary
- Remove `Sudo.sudo` wrapping from `authorizeAccount()`, `authorizePreimage()`, and `storeProof()` in all JS examples
- Rename `sudoSigner` → `authorizationSigner` across all example files
- Alice is already configured as a `TestAccount` via `EnsureSignedBy<TestAccounts>` in both runtimes, so she can sign authorization calls directly without sudo elevation

Closes #288

## Test plan
- [ ] Run `authorize_and_store_papi.js` against a local dev node
- [ ] Run `authorize_preimage_and_store_papi.js` against a local dev node
- [ ] Run `store_chunked_data.js` against a local dev node
- [ ] Verify authorization succeeds without Sudo wrapping

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/console-ui/src/pages/Authorizations/Authorizations.tsx
+++ b/console-ui/src/pages/Authorizations/Authorizations.tsx
@@ -404,34 +404,12 @@ function FaucetAuthorizePreimagePanel() {
         max_size: sizeValue > 0n ? sizeValue : 1024n * 1024n,
       });
 
-      await new Promise<void>((resolve, reject) => {
-        let resolved = false;
-
-        const subscription = tx.signSubmitAndWatch(aliceSigner).subscribe({
-          next: (ev: any) => {
-            console.log("TX event:", ev.type);
-            if (ev.type === "txBestBlocksState" && ev.found && !resolved) {
-              resolved = true;
-              subscription.unsubscribe();
-              resolve();
-            }
-          },
-          error: (err: any) => {
-            if (!resolved) {
-              resolved = true;
-              reject(err);
-            }
-          },
-        });
-
-        setTimeout(() => {
-          if (!resolved) {
-            resolved = true;
-            subscription.unsubscribe();
-            reject(new Error("Transaction timed out"));
-          }
-        }, 120000);
-      });
+      // Wait for finalization (not just best block inclusion) so that
+      // subsequent storage queries can see the new authorization.
+      const result = await tx.signAndSubmit(aliceSigner);
+      if (!result.ok) {
+        throw new Error("Transaction dispatch failed");
+      }
 
       setSubmitSuccess("Successfully authorized preimage");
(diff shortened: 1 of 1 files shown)


examples/store_chunked_data.js
```diff
@@ -147,11 +147,7 @@ async function storeProof(typedApi, sudoSigner, whoSigner, rootCID, dagFileBytes
     // This can be a serious pallet, this is just a demonstration.
     const proof = `ProofCid: ${rawDagCid.toString()} -> rootCID: ${rootCID.toString()}`;
     const remarkTx = typedApi.tx.System.remark({ remark: Binary.fromText(proof) });
-    const sudoTx = typedApi.tx.Sudo.sudo({ call: remarkTx.decodedCall });
-    await sudoTx.signSubmitAndWatch(sudoSigner).subscribe({
-        next: (ev) => console.log(`✅ Proof remark event:`, ev.type),
-        error: (err) => console.error(`❌ Proof remark error:`, err),
-    });
+    await waitForTransaction(remarkTx, whoSigner, "StoreProofRemark");
```
x3c41a: Ok, so `system.remark` fails because it requires `sudo`. Though, `remark` is used just for demonstration so we can remove it. @bkontur , shall I go ahead and remove this operation or you want me to revert it back to `sudo.` call?

P.S.
The bulletin-polkadot runtime has a strict ValidateSigned call filter that whitelists only specific calls:
  - TransactionStorage calls
  - Session::set_keys / Session::purge_keys
  - Bridge calls
  - Proxy, Sudo, Utility
  - System::apply_authorized_upgrade (only this one System call)

Everything else, including  System::remark, is rejected with InvalidTransaction::Call

examples/store_chunked_data.js
```diff
@@ -204,7 +200,7 @@ async function main() {
         const { rootCid, dagBytes } = await buildUnixFSDag(metadataJson, 0xb220)
 
         // Store DAG and proof to the Bulletin.
-        let { rawDagCid } = await storeProof(bulletinAPI, sudoSigner, whoSigner, rootCid, Buffer.from(dagBytes));
+        let { rawDagCid } = await storeProof(bulletinAPI, whoSigner, rootCid, Buffer.from(dagBytes));
```
bkontur:
```suggestion
        let { rawDagCid } = await storeProof(bulletinAPI, authorizationSigner, rootCid, Buffer.from(dagBytes));
```

PR #241, opened 2026-02-12
Node sync modes tests based on zombienet-sdk

karolk91:
Adds a new `zombienet-sdk-tests` crate with integration tests that spawn local
  parachain networks via `zombienet-sdk` and verify sync modes plus
  transaction-storage (bitswap) behavior on the bulletin parachain.

  All tests are gated behind the `zombie-sync-tests` feature flag so they are
  not picked up by `cargo test --workspace`.

  ## Test matrix (9 parachain tests)

  | Test | Sync mode | Pruning | Expected outcome |
  |---|---|---|---|
  | `parachain_fast_sync_test` | fast | no | Sync completes, bitswap returns `DONT_HAVE` |
  | `parachain_fast_sync_with_pruning_test` | fast | yes | Sync fails (pruned blocks) |
  | `parachain_warp_sync_test` | warp | no | Sync completes, bitswap returns `DONT_HAVE` |
  | `parachain_warp_sync_with_pruning_test` | warp | yes | Sync completes, bitswap returns `DONT_HAVE` |
  | `parachain_full_sync_test` | full | no | Sync completes, bitswap serves data |
  | `parachain_full_sync_with_pruning_test` | full | yes | Sync fails (pruned blocks) |
  | `parachain_full_sync_relay_warp_sync_test` | full (para) + warp (relay) | no | Parachain downloads indexed bodies, bitswap serves data |
  | `parachain_rpc_node_bitswap_test` | full | no | Collator stores → full node syncs → RPC node serves via bitswap |
  | `parachain_ldb_storage_verification_test` | – | yes | Verifies RocksDB `col11` refcounting and data expiration after retention |

  ## Behaviors verified

  - **Full sync** downloads indexed transaction bodies — synced nodes serve data via bitswap.
  - **Fast sync** skips block bodies — synced nodes return `DONT_HAVE`.
  - **Warp sync** gap-fill downloads block bodies but does not execute them
    (BODY column, not TRANSACTIONS) — synced nodes return `DONT_HAVE`.
  - **Block pruning** blocks fast/full sync when all peers prune historical data.
  - **Warp sync with pruning** still completes (uses GRANDPA warp proofs).
  - **LDB tests** verify `col11` refcounts on duplicate stores and data
    expiration after the retention period.
  - **Multi-hop bitswap** propagates from collator → full node → RPC node.

  ## Crate layout
  
    - `Cargo.toml`, `build.rs` — crate manifest + bitswap protobuf codegen
[...]

PR #313, opened 2026-03-05
renew storage by content_hash & auto renewal

rosarp:
## Summary

  - Adds **`TransactionByContentHash`** storage map enabling renewal addressed by content hash instead of `(block, index)`.
  - Adds **`renew_content_hash(content_hash)`** extrinsic.
  - Implements **automatic data renewal** via:
    - `enable_auto_renew(content_hash)` / `disable_auto_renew(content_hash)` extrinsics.
    - A **composite mandatory inherent** `apply_block_inherents(Option<TransactionStorageProof>)` that **replaces the standalone `check_proof` extrinsic** and additionally drains `PendingAutoRenewals`.
  - Promotes `renew` and `renew_content_hash` to `DispatchClass::Operational`.

  ## `renew_content_hash`

  - `TransactionByContentHash: StorageMap<ContentHash, (BlockNumber, ChunkIndex)>` tracks the most recent location for each content hash.
  - Populated by `store` / `store_with_cid_config`, overwritten on every renewal.
  - Cleaned up in `on_initialize` only when the entry still points at the obsolete block — renewed entries' mappings survive.
  - Both signed and unsigned validation paths recognise the new call.

  ## Auto-renewal system

  ### Lifecycle
  1. User calls `enable_auto_renew(hash)`. Validation requires the hash to be currently stored and the caller to have an active authorization with enough remaining quota for at least one renewal. The opt-in is recorded in `AutoRenewals[hash] = AutoRenewalData { account }`.
  2. At the next expiry block:
     - `on_initialize` consumes `Transactions[obsolete]` and, for every entry with a registered `AutoRenewals[hash]`, pushes `(hash, tx_info,
  renewal_data)` onto `PendingAutoRenewals`.
     - The mandatory `apply_block_inherents` inherent drains the queue in `do_process_auto_renewals`. Per item:
       - `check_authorization(scope, size, consume=true, is_renew=true)` enforces the per-account and chain-wide caps (defined by main's authorization model).
       - On success: re-indexes the data via `transaction_index::renew`, appends a fresh `TransactionInfo { kind: Renew, … }`, overwrites `TransactionByContentHash[hash]`, emits `DataAutoRenewed`.
       - On failure (auth gone, quota exceeded, or `BlockTransactions` full): **removes the registration**, emits `AutoRenewalFailed`, lets the data expire. Other items in the same batch are unaffected.
  3. `on_finalize` asserts `PendingAutoRenewals` is empty and `ProofChecked` is set, guaranteeing block authors include the inherent.

  ### Storage
  - `AutoRenewals: StorageMap<ContentHash, AutoRenewalData<AccountId>>` — persistent opt-in.
  - `PendingAutoRenewals: StorageValue<BoundedVec<(ContentHash, TransactionInfo, AutoRenewalData), MaxBlockTransactions>>` — transient, populated and drained within the same block.

  ### Events / errors
  - New events: `AutoRenewalEnabled`, `AutoRenewalDisabled`, `DataAutoRenewed`, `AutoRenewalFailed`.
  - New errors: `AutoRenewalAlreadyEnabled`, `AutoRenewalNotEnabled`, `NotAutoRenewalOwner`.

  ### Composite mandatory inherent
  - `check_proof` is removed; both the proof verification and the auto-renewal drain are now handled by `apply_block_inherents(proof:
  Option<TransactionStorageProof>)` so block authors satisfy both `on_finalize` invariants atomically.
  - `is_inherent` / `ValidateUnsigned` updated to recognise the new dispatchable.
  - Proof verification uses `transactions_at(target)` so historical entries from before main's v2→v3 `TransactionInfo` migration still verify  while that multi-block migration is in flight.


examples/check_auto_renew_chopsticks.js
```diff
@@ -0,0 +1,339 @@
...
+  if (expiryTriggered) {
+    logOk("AUTO-RENEWAL TEST PASSED:");
+    logOk("  1. Base storage state set correctly (authorization, transactions, content hash map)");
+    logOk("  2. enable_auto_renew: AutoRenewals entry created and verified");
+    logOk("  3. disable_auto_renew: AutoRenewals entry removed and verified");
+    logOk("  4. Auto-renewal re-enabled for expiry test");
+    logOk("  5. on_initialize at the expiry block found the expiring data");
+    logOk("  6. AutoRenewals registration was found for the content hash");
+    logOk("  7. PendingAutoRenewals was populated (proven by on_finalize panic)");
+    logOk("  8. on_finalize correctly asserted that process_auto_renewals must run");
```
franciscoaguirre: This are good checks but I would also include the inherent and check that on_finalize doesn't panic. For this we would need our own block authors, so probably chopsticks is better for this whole test? We don't gain much from forking the live chain state anyway, do we?

pallets/transaction-storage/src/lib.rs
```diff
@@ -296,6 +296,12 @@ pub mod pallet {
 		AuthorizationNotExpired,
 		/// Content hash was not calculated.
 		InvalidContentHash,
+		/// Auto-renewal is already enabled for this content hash.
+		AutoRenewAlreadyEnabled,
+		/// Auto-renewal is not enabled for this content hash.
+		AutoRenewNotEnabled,
+		/// Caller is not the owner of the auto-renewal registration.
+		NotAutoRenewOwner,
```
franciscoaguirre:
```suggestion
		/// Auto-renewal is already enabled for this content hash.
		AutoRenewalAlreadyEnabled,
		/// Auto-renewal is not enabled for this content hash.
		AutoRenewalNotEnabled,
		/// Caller is not the owner of the auto-renewal registration.
		NotAutoRenewalOwner,
```

pallets/transaction-storage/src/lib.rs
```diff
@@ -692,6 +759,188 @@ pub mod pallet {
...
+				Error::<T>::RenewedNotFound
+			);
+			ensure!(
+				!AutoRenewals::<T>::contains_key(content_hash),
+				Error::<T>::AutoRenewAlreadyEnabled
+			);
+
+			// Verify the account has authorization sufficient for at least one future renewal.
+			let (block, index) = TransactionByContentHash::<T>::get(content_hash)
+				.ok_or(Error::<T>::RenewedNotFound)?;
```
franciscoaguirre: The get().ok_or()? already does the check that you're doing above with contains_key(), I would just keep the get()

pallets/transaction-storage/src/lib.rs
```diff
@@ -692,6 +759,188 @@ pub mod pallet {
...
+								cid_codec: tx_info.cid_codec,
+								block_chunks: total_chunks,
+							})
+							.map_err(|_| Error::<T>::TooManyTransactions)
+					})?;
+
+					TransactionByContentHash::<T>::insert(
+						content_hash,
+						(frame_system::Pallet::<T>::block_number(), new_index),
+					);
```
franciscoaguirre: This logic seems duplicated in renew(), renew_by_content_hash() and here. Create a do_renew() function to deduplicate

franciscoaguirre (commented):
renew_content_hash looks good and will improve user experience greatly. I added some comments around auto renewal that make me worried about it

pallets/transaction-storage/src/lib.rs
```diff
@@ -692,6 +733,146 @@ pub mod pallet {
...
+		/// data size). The authorization is **not** consumed here; it is consumed each time
+		/// the data is auto-renewed (every `StoragePeriod` blocks).
+		/// Authorization is checked here but might still be missing when actually renewed.
+		///
+		/// Emits [`AutoRenewalEnabled`](Event::AutoRenewalEnabled) when successful.
+		#[pallet::call_index(12)]
+		#[pallet::weight(T::WeightInfo::enable_auto_renew())]
+		pub fn enable_auto_renew(
+			origin: OriginFor<T>,
+			content_hash: ContentHash,
```
bkontur, line 844: @karolk91 @rosarp I am just thinking about  - for 1 SPA with 10 chunks, what would be better:
- 11x enable_auto_renew or
- 1x enable_auto_renew

```suggestion
			content_hash: Vec<ContentHash>,
```

pallets/transaction-storage/src/lib.rs
```diff
@@ -840,6 +1053,12 @@ pub mod pallet {
 		type Call = Call<T>;
 
 		fn validate_unsigned(_source: TransactionSource, call: &Self::Call) -> TransactionValidity {
+			// Inherent-style calls (check_proof, process_auto_renewals) are injected by
+			// the block author, not the transaction pool. Return a valid but empty
+			// transaction if one arrives here.
+			if Self::is_inherent(call) {
+				return Ok(ValidTransaction::default());
+			}
```
bkontur: @rosarp why do we need this here? it is part of the `fn pre_dispatch` already

pallets/transaction-storage/src/lib.rs
```diff
@@ -692,6 +733,146 @@ pub mod pallet {
...
+		/// account no longer has sufficient authorization the renewal is skipped (not panicked)
+		/// and an [`AutoRenewalFailed`](Event::AutoRenewalFailed) event is emitted. Auto-renewal
+		/// registration is automatically removed for failed items.
+		#[pallet::call_index(11)]
+		#[pallet::weight((T::WeightInfo::process_auto_renewals(T::MaxBlockTransactions::get()), DispatchClass::Mandatory))]
+		pub fn process_auto_renewals(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
+			ensure_none(origin)?;
+			let pending = PendingAutoRenewals::<T>::take();
+
+			for (content_hash, tx_info, renewal_data) in pending.into_iter() {
```
bkontur: this does not compile?

```suggestion
			for (content_hash, tx_info, renewal_data) in PendingAutoRenewals::<T>::take() {
```

examples/check_auto_renew_chopsticks.js
```diff
@@ -0,0 +1,339 @@
...
+  // TransactionInfo { chunk_root: H256, content_hash: [u8;32], hashing: HashingAlgorithm(u8 enum), cid_codec: CidCodec(u64), size: u32, block_chunks: u32 }
+  let data = "04"; // compact len = 1
+  data += chunkRoot; // chunk_root
+  data += toHex(contentHash); // content_hash
+  data += "00"; // hashing = Blake2b256 (variant 0)
+  data += encodeU64LE(0x55); // cid_codec = RAW_CODEC
+  data += encodeU32LE(size); // size
+  data += encodeU32LE(chunks); // block_chunks
+  return "0x" + data;
+}
```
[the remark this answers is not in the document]
rosarp: rewrote with papi

pallets/transaction-storage/src/lib.rs
kianenigma: This is an important follow-up -- the manual DB weights not only ignore execution ref-time, they also ignore the PoV part of the storage operations. Look into what `T::DbWeight::get();` actually returns -- it is just `ref_time`.

kianenigma: On top of this: on a parachain we should always prefer to use on_poll, if it is the case.

pallets/transaction-storage/src/lib.rs
```diff
@@ -692,6 +733,146 @@ pub mod pallet {
...
+
+			let new_index = Self::do_renew(info)?;
+			Self::deposit_event(Event::Renewed { index: new_index, content_hash });
+			Ok(().into())
+		}
+
+		/// Process all pending auto-renewals for this block.
+		///
+		/// This is a **mandatory** extrinsic: block authors must include it whenever
+		/// `on_initialize` populated [`PendingAutoRenewals`]. Failure to include it will
```
kianenigma: Is this flow validated such that the weight of work that on-init does + the weight of the inherent will not exceed the block weight limits? specifically, since this will run as a parachain, the PoV size.

pallets/transaction-storage/src/lib.rs
```diff
@@ -692,6 +733,146 @@ pub mod pallet {
...
+		/// of the original data, as emitted in the [`Stored`](Event::Stored) or
+		/// [`Renewed`](Event::Renewed) event.
+		///
+		/// This is a convenience alternative to [`renew`](Self::renew) that does not require
+		/// knowing the exact `(block_number, tx_index)` pair.
+		///
+		/// Emits [`Renewed`](Event::Renewed) when successful.
+		#[pallet::call_index(10)]
+		#[pallet::weight((T::WeightInfo::renew_content_hash(), DispatchClass::Operational))]
+		pub fn renew_content_hash(
```
skunert, line 814: This needs to be added to the call filter to disallow batch and other wrappers here https://github.com/paritytech/polkadot-bulletin-chain/blob/main/pallets/transaction-storage/src/extension.rs#L69-L69

skunert: And probably [here](https://github.com/paritytech/polkadot-bulletin-chain/blob/main/pallets/transaction-storage/src/extension.rs#L207-L207) too?

pallets/transaction-storage/src/lib.rs
```diff
@@ -831,7 +1044,7 @@ pub mod pallet {
 		}
 
 		fn is_inherent(call: &Self::Call) -> bool {
-			matches!(call, Call::check_proof { .. })
+			matches!(call, Call::check_proof { .. } | Call::process_auto_renewals { .. })
```
skunert: How is it enforced that this inherent is actually included in a block? This needs to be included in the above create_inherent or be its own inherent. Otherwise block producer will not include this and the chain will stall.

pallets/transaction-storage/src/lib.rs
```diff
@@ -944,6 +1167,44 @@ pub mod pallet {
...
+		/// [`process_auto_renewals`](Self::process_auto_renewals).
+		///
+		/// Indexes the renewal via `sp_io::transaction_index::renew`, pushes a new
+		/// `TransactionInfo` into [`BlockTransactions`], and updates
+		/// [`TransactionByContentHash`]. Returns the new transaction index on success.
+		fn do_renew(info: TransactionInfo) -> Result<u32, Error<T>> {
+			let extrinsic_index =
+				<frame_system::Pallet<T>>::extrinsic_index().ok_or(Error::<T>::BadContext)?;
+			let content_hash = info.content_hash;
+			sp_io::transaction_index::renew(extrinsic_index, content_hash);
```
skunert, line 1402: I think we should move this hostcall after the mutation below. If try_push fails with `TooManyTransactions`, the call here will still have caused side effects on the node side and wrongly increase the ref count. Maybe even more fallout, did not check.

pallets/transaction-storage/src/lib.rs
rosarp: This will be fixed by either of these PRs:
- (Internally called both check_proof & process_auto_renewals) https://github.com/rosarp/polkadot-bulletin-chain/pull/4/changes
- (Separated in separate pallet) https://github.com/rosarp/polkadot-bulletin-chain/pull/3

pallets/transaction-storage/src/lib.rs
karolk91: thanks @skunert for flagging, this case should be not possible to reach assuming that the tx validation before dispatch is implemented correctly. At the same time I don't see a reason not to move it a bit further so just did it

pallets/transaction-storage/src/lib.rs
karolk91: it is now implemented as `apply_block_inherents` that does both `check_proof` and `process_auto_renewals`

pallets/transaction-storage/src/lib.rs
karolk91: thats correct, added to both places

pallets/transaction-storage/src/lib.rs
karolk91: as of today, there are two benchmarks added for `apply_block_inherents` and `on_initialize_with_expiry`, and there is also `ensure_weight_sanity` that takes these into account

In my practical tests, the worst case scenario results in around 3-4% for both ref time and proof size in a block

pallets/transaction-storage/src/lib.rs
karolk91: we use `on_initialize` because we want the guarantee that the logic will be executed when time comes - one block later and data would be gone from state

pallets/transaction-storage/src/benchmarking.rs
```diff
@@ -381,5 +395,106 @@ mod benchmarks {
...
+
+		#[extrinsic_call]
+		_(RawOrigin::Signed(caller.clone()), content_hash);
+
+		assert_last_event::<T>(Event::AutoRenewDisabled { content_hash, who: caller }.into());
+		Ok(())
+	}
+
+	#[benchmark]
+	fn process_auto_renewals() -> Result<(), BenchmarkError> {
```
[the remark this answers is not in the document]
karolk91: benchmarking for this is now implemented as [apply_block_inherents](https://github.com/paritytech/polkadot-bulletin-chain/pull/313/files#diff-e7798fc7fce74143e0eeb7d72ea814b0126d299d20ab53c0672e0d3ae1832536R434)

pallets/transaction-storage/src/lib.rs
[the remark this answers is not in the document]
karolk91: it is now the role of [`Call::apply_block_inherents`](https://github.com/paritytech/polkadot-bulletin-chain/pull/313/files#diff-267ebd494899c5554d9a47332f40a3fd4d73eef822ce0eb9682f8ff92f18187cR1043)

pallets/transaction-storage/src/tests.rs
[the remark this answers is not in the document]
karolk91: covered by [`auto_renew_permissionless_transfer`](https://github.com/paritytech/polkadot-bulletin-chain/pull/313/files#diff-0502a5245c151b9b5d78f6760c9a317d1b928793c3069faa3b48d6c5a8a26863R1963)

examples/check_auto_renew_papi.js
```diff
@@ -0,0 +1,339 @@
...
+  if (expiryTriggered) {
+    logOk("AUTO-RENEWAL TEST PASSED:");
+    logOk("  1. Base storage state set correctly (authorization, transactions, content hash map)");
+    logOk("  2. enable_auto_renew: AutoRenewals entry created and verified");
+    logOk("  3. disable_auto_renew: AutoRenewals entry removed and verified");
+    logOk("  4. Auto-renewal re-enabled for expiry test");
+    logOk("  5. on_initialize at the expiry block found the expiring data");
+    logOk("  6. AutoRenewals registration was found for the content hash");
+    logOk("  7. PendingAutoRenewals was populated (proven by on_finalize panic)");
+    logOk("  8. on_finalize correctly asserted that process_auto_renewals must run");
```
karolk91: there are plenty of tests based on zombienet-sdk - but these also depend on https://github.com/paritytech/polkadot-bulletin-chain/pull/241. I would like not to merge into this PR here, as its a lot of code - I will publish a PR separately for these

pallets/transaction-storage/src/lib.rs
```diff
@@ -692,6 +759,188 @@ pub mod pallet {
...
+						transactions
+							.try_push(TransactionInfo {
+								chunk_root: tx_info.chunk_root,
+								size: tx_info.size,
+								content_hash: tx_info.content_hash,
+								hashing: tx_info.hashing,
+								cid_codec: tx_info.cid_codec,
+								block_chunks: total_chunks,
+							})
+							.map_err(|_| Error::<T>::TooManyTransactions)
```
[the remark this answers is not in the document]
karolk91: I believe this is now covered by [process_auto_renewals_continues_on_per_item_failure](https://github.com/paritytech/polkadot-bulletin-chain/pull/313/files#diff-0502a5245c151b9b5d78f6760c9a317d1b928793c3069faa3b48d6c5a8a26863R2015)

pallets/transaction-storage/src/lib.rs
karolk91: Not sure if I understand the comment right but `validate_unsigned` and `pre_dispatch` are executed at different stages (pool admission vs block production)

pallets/transaction-storage/src/benchmarking.rs
```diff
@@ -275,6 +252,10 @@ mod benchmarks {
 		TransactionStorage::<T>::authorize_account(origin, who.clone(), 0, 1)
 			.map_err(|_| BenchmarkError::Stop("unable to authorize account"))?;
 
+		// `AuthorizationPeriod` is ~90 days of blocks on real runtimes; iterating
```
franciscoaguirre:
```suggestion
		// `AuthorizationPeriod` is ~14 days of blocks on real runtimes; iterating
```

It's only 90 days in test runtimes for now

pallets/transaction-storage/src/tests.rs
```diff
@@ -1492,6 +1583,624 @@ fn authorize_storage_extension_passes_through_non_storage_calls() {
...
+		let data = vec![0u8; 2000];
+		let content_hash = blake2_256(&data);
+
+		// Authorize with exactly enough for 2 operations (store doesn't consume here,
+		// since it's unsigned, but renew does via process_auto_renewals)
+		assert_ok!(TransactionStorage::authorize_account(RuntimeOrigin::root(), who, 3, 6000));
+		assert_ok!(TransactionStorage::store(RuntimeOrigin::none(), data));
+		run_to_block(2, || None);
+		assert_ok!(
+			TransactionStorage::enable_auto_renew(RuntimeOrigin::signed(who), content_hash,)
```
franciscoaguirre, line 2036: As an optimization we could add a `store_and_auto_renew()` call that doesn't consume the allowance twice. We can do something like that in a future PR

pallets/transaction-storage/src/lib.rs
```diff
@@ -362,16 +373,44 @@ pub mod pallet {
...
+						// renewed elsewhere and points at a different block).
+						if let Some((block, _)) = TransactionByContentHash::<T>::get(hash) {
+							if block == obsolete {
+								TransactionByContentHash::<T>::remove(hash);
+							}
+						}
+						// If auto-renewal is registered for this hash, schedule it for the
+						// inherent to drain. `try_push` silently drops items beyond
+						// MaxBlockTransactions — auto-renewal is best-effort; excess items
+						// simply won't be renewed this block.
```
franciscoaguirre: Auto-renewal is not best-effort. The thing is `|PendingAutoRenewals| <= MaxBlockTransactions` so this will never error

pallets/transaction-storage/src/lib.rs
```diff
@@ -796,6 +925,30 @@ pub mod pallet {
...
+	#[pallet::storage]
+	pub type AutoRenewals<T: Config> =
+		StorageMap<_, Blake2_128Concat, ContentHash, AutoRenewalData<T::AccountId>, OptionQuery>;
+
+	/// Transactions that must be auto-renewed in the current block.
+	///
+	/// Populated by `on_initialize` when a block's data is about to expire.
+	/// Cleared by the `apply_block_inherents` mandatory inherent executed in the same block.
+	#[pallet::storage]
+	pub(super) type PendingAutoRenewals<T: Config> = StorageValue<
```
franciscoaguirre, line 1019: We should add a config item of how many transactions in the block can enable auto renew, right?

franciscoaguirre (approved):
Looks good to me 🚀  We can address feedback in follow-up PRs

karolk91 (approved):

pallets/transaction-storage/src/lib.rs
karolk91: It is currently bounded by `MaxBlockTransactions` or you mean a separate limit ?

pallets/transaction-storage/src/tests.rs
karolk91: Yes, thats definitely some idea that circulated around - together with multiple auto-renewal slots per item

pallets/transaction-storage/src/lib.rs
```diff
@@ -362,16 +373,44 @@ pub mod pallet {
...
 			// Drop obsolete roots. The proof for `obsolete` will be checked later
 			// in this block, so we drop `obsolete` - 1.
-			weight.saturating_accrue(db_weight.reads(1));
 			let period = Self::retention_period();
 			let obsolete = n.saturating_sub(period.saturating_add(One::one()));
+			let mut num_expiring: u32 = 0;
 			if obsolete > Zero::zero() {
-				weight.saturating_accrue(db_weight.writes(1));
-				<Transactions<T>>::remove(obsolete);
+				if let Some(transactions) = <Transactions<T>>::take(obsolete) {
```
bkontur, line 434: @karolk91 @rosarp this part looks risky, `on_initialize` should be pretty fast and slim, and here we are doing lots of storage actions:
```
1x Transactions::take (decoding max 512txs)
1x PendingAutoRenewals::get
Max 512x TransactionByContentHash::<T>::get(hash)
Max 512x TransactionByContentHash::<T>::remove(hash);
AutoRenewals::get
PendingAutoRenewals::put
```

Also Cisco wants to read all the transactions `on_initialize` here: https://github.com/paritytech/polkadot-bulletin-chain/pull/486/files#diff-267ebd494899c5554d9a47332f40a3fd4d73eef822ce0eb9682f8ff92f18187cR420

---
@karolk91 @franciscoaguirre  Are we sure that this won't stuck the chain in maximum usage?
What about doing some lazy feature? `Transactions::take(obsolete)` will just copy data to some processing queue, e.g.
```
pub(super) type DeferredTransactions<T: Config> = StorageMapOrValueOrWhatever< // <- maybe does not need key just some list
```
and then we process this lazy `on_poll`, I did something similar here: https://github.com/paritytech/polkadot-bulletin-chain/pull/457

karolk91: @bkontur, specifically for auto-renew:
1. We need to take into account the block pruning on the node side - whatever mechanism we figure out - it must execute before block is pruned. There is no such guarantee with `on_poll`. Inherents are also executed before `on_poll` so the mechanism running there would need to "schedule" inherents for future blocks
3. Benchmarking code is critical here - If I got it right, currently the worst case scenario consumes about 4% of block (proof size and ref time)

scripts/cmd/cmd.py
```diff
@@ -179,7 +179,8 @@ def setup_logging():
                                f"--steps={args.steps} "
                                f"--repeat={args.repeat} "
                                f"--heap-pages=4096 "
-                               f"--min-duration 1 "
```
bkontur, line 182: @karolk91 why `f"--min-duration 1 "` removed? I think we need this with stable2603 bencher - https://github.com/polkadot-fellows/runtimes/issues/1130#issuecomment-4205581984

pallets/transaction-storage/src/lib.rs
bkontur: > 2. Benchmarking code is critical here - If I got it right, currently the worst case scenario consumes about 4% of block (proof size and ref time)

Ok, so how could we ensure that `on_initialize` is 100% correct? Can we add some sanity weights checks:
- for some block we push 512 transactions
- for some block we push 8x 1MiB + 504 small tx 
- we schedule renew for all, so the some `on_initialize` will hit the all the storage operations

If we add:
- sanity weights test to the pallet
- runtime level tests that ensures above

scripts/cmd/cmd.py
karolk91: With these changes I've followed the polkadot-sdk configs - but I now reverted this to match fellows repo

pallets/transaction-storage/src/lib.rs
karolk91: We have [`ensure_weight_sanity` test](https://github.com/paritytech/polkadot-bulletin-chain/pull/313/files#diff-267ebd494899c5554d9a47332f40a3fd4d73eef822ce0eb9682f8ff92f18187cR1966) and I've also zombienet-sdk based tests that I didn't want to be part of this PR as these are based on https://github.com/paritytech/polkadot-bulletin-chain/pull/241 and would bring a lot of unrelated changes

examples/justfile
```diff
@@ -645,6 +645,31 @@ run-test-authorize-preimage-and-store test_dir ws_url="ws://127.0.0.1:10000" see
...
     node authorize_preimage_and_store_papi.js "{{ ws_url }}" "{{ seed }}" "{{ http_ipfs_api }}"
 
+# Run auto-renewal Chopsticks test
+# Tests enable/disable auto-renewal and verifies on_initialize scheduling.
+# Builds the Westend runtime WASM by default so auto-renewal code is included.
+# Parameters:
+#   endpoint - WebSocket URL of the Bulletin chain (default: wss://westend-bulletin-rpc.polkadot.io)
+#   wasm_path - Path to runtime WASM override (default: auto-built from source)
+#   skip_build - Set to "true" to skip WASM build (default: false)
+run-test-auto-renew-chopsticks endpoint="wss://westend-bulletin-rpc.polkadot.io" wasm_path="" skip_build="false": npm-install
```
bkontur, line 655: @karolk91 can we make this part of the integration-tests.yml CI?

karolk91: yes, will do that in separate PR as it seems that this script has some issues to run properly

karolk91:
Block with 512 auto renewals:
image

auto-renewal fail when limit per account reached:
image

PR #344, opened 2026-03-20
Replace sp_io::trie host functions with sp_trie equivalents

### LINKED ISSUE #341: Chopsticks doesn't work for blocks where proofs are verified
The goal of the issue is to investigate the error below which happens when chopsticks is trying to build a new block that should contain a `transactionStorage.checkProof` inherent:

`Error: Host function not implemented: ext_trie_blake2_256_verify_proof_version_2`

Path to reproduce (the block 637328 used here is just some random block that has `checkProof` inherent):

`npx @acala-network/chopsticks@latest run-block --config=configs/paseo-bulletin-override.yaml --block 637328`

where `configs/paseo-bulletin-override.yaml`:

```
endpoint: wss://paseo-bulletin-rpc.polkadot.io
mock-signature-host: true
db: ./db_bulletin_paseo.sqlite
runtime-log-level: 5
```

x3c41a:
## Summary

- Replace `sp_io::trie::blake2_256_ordered_root` and `sp_io::trie::blake2_256_verify_proof` host function calls in `pallet-transaction-storage` with direct `sp_trie` pure-Rust equivalents
- This makes the runtime compatible with Chopsticks, which doesn't implement `ext_trie_blake2_256_verify_proof_version_2`

Closes #341

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9109,6 +9109,7 @@ dependencies = [
  "polkadot-sdk-frame",
  "scale-info",
  "sp-transaction-storage-proof",
+ "sp-trie",
  "tracing",
  "transaction-storage-primitives",
 ]
--- a/pallets/transaction-storage/Cargo.toml
+++ b/pallets/transaction-storage/Cargo.toml
@@ -22,6 +22,7 @@ transaction-storage-primitives = { workspace = true }
 polkadot-sdk-frame = { workspace = true, default-features = false, features = [
 	"runtime",
 ] }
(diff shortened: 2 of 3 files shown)


bkontur:
> @x3c41a, this change seem to move the calculations from the native/host execution to the WASM code - avoiding the host calls that are not implemented by smoldot.

Hmm, let's be very sure about merging this change. I don't know what are the trade-offs here, but what about just adding those host functions to the smoldot like we did here: https://github.com/smol-dot/smoldot/pull/2189 ?

PR #348, opened 2026-03-24
Add HOP promotion: runtime-side

franciscoaguirre:
**Should be compatible with** https://github.com/paritytech/polkadot-sdk/pull/11662

---

In order to support promotion to Bulletin in HOP, we need both node-side and runtime-side changes.
This PR adds the runtime-side changes in order to test them before we [move Bulletin Polkadot to fellowships](https://github.com/bkontur/runtimes/pull/23).
The node-side changes are in https://github.com/paritytech/polkadot-sdk/pull/11484.

## How promotion works

The node regularly scans for entries about to expire and for each, it calls the runtime API HopPromotionAPI to get the promote() extrinsic it needs from pallet-hop-promotion. This extrinsic is authorized only if it comes from the local transaction pool, meaning it can't be sent from the outside world. It can only be used by collators building a block to promote. Since the extrinsic is also given a low priority from the runtime, it'll only be included in a block if it has space available after all user extrinsics.

sp-hop was created here mimicking the same crate on the sdk until we merge that PR.

## Approach

The promote() extrinsic:
- Uses general transactions via FRAME's #[pallet::authorize] macro
- Call checks the user signature, collator stores the signature since hop_submit until the time of promotion
- No fees. Protocol-level operation not submittable by regular users
- Priority 0, the minimum. To make sure all user transactions take precedence over these ones.
- Local-only, TransactionSource::External gets rejected.
- Not propagated to other nodes
- Longevity 5, might be small. It will however be constantly re-created if needed by the node long-running task.
- Delegates to pallet_transaction_storage::Pallet::do_store(data, Blake2b256, RAW_CODEC)

## Companion: sdk

The companion to this PR also declares the runtime API and uses that to submit these promote() extrinsics.


pallets/hop-promotion/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+		#[pallet::call_index(0)]
+		#[pallet::weight(
+			<T as pallet_transaction_storage::Config>::WeightInfo::store(data.len() as u32)
+		)]
+		#[pallet::authorize(|source, data: &Vec<u8>| {
+			if matches!(source, TransactionSource::External) {
+				return Err(InvalidTransaction::Call.into());
+			}
+			if data.is_empty() ||
+				data.len() >
```
bkontur: @franciscoaguirre We have dedicated `ensure_data_size_ok` or `data_size_ok` please try reuse those somehow (Self:: should work)

bkontur (approved):

PR #379, opened 2026-04-02
Delete more artifacts and add reference to networks.ts

RafalMirowski1:


RafalMirowski1:
also included https://github.com/paritytech/polkadot-bulletin-chain/issues/400 here.
- Move polkadot-api and multiformats to peer dependencies
- Remove hardcoded VERSION export
- Remove "types": ["node"] from tsconfig.json

https://github.com/paritytech/polkadot-bulletin-chain/issues/400 mentions also to add coverage to .gitignore (of sdk) but was added to global .gitignore (and contents of coverage are removed here)

RafalMirowski1:
@bkontur 
> > > @RafalMirowski1 please, also change "Test Rust SDK" test to use waiting for "in block" instead of "finalized" https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/23899037762/job/69691035038?pr=364 (to speed up CI)
> > 
> > 
> > I'm not sure rust sdk exposes this option
> 
> hmm, I think it should, does TS SDK support this? @karolk91

ts sdk has some `WaitFor` object eg `WaitFor.InBlock`/`WaitFor.Finalized`. Rust test in ci uses `store` and `store_with_progress` which explicitly loop until `TxStatus::InFinalizedBlock`, and `authorize_account` which calls `submit_and_finalize` under the hood, there is `wait_for_finalization` in `StoreOptions` but it doesnt seem to anything here.

bkontur:
> `submit_and_finalize` under the hood, there is `wait_for_finalization` in `StoreOptions` but it doesnt seem to anything here.

@RafalMirowski1 what does this mean? That rust sdk does not support "in block" just finalized? If so, please, follow up to align with TS SDK

bkontur (approved):
@RafalMirowski1 can we close now https://github.com/paritytech/polkadot-bulletin-chain/issues/400?

PR #389, opened 2026-04-07
Fix flaky vitest integration tests

x3c41a:
Fixes #385

## Summary

The TS SDK integration tests timed out intermittently in CI because the per-transaction timeout was hardcoded to 120s with no way to configure it.

## Changes

- Add `txTimeout` to `ClientConfig` (default: 120s), replacing the hardcoded value
- Extract `DEFAULT_CLIENT_CONFIG` to deduplicate defaults across `AsyncBulletinClient`, `MockBulletinClient`, and `BulletinPreparer`
- Rename internal `PapiEventMappingResult` to `MappedTxStatus` for clarity
- Set `txTimeout: 180_000` in integration tests to handle slow CI runners
- Add per-test `timeout: 180_000` on chunked upload tests (multiple sequential transactions)

## Test plan

- [x] `npm run lint` passes
- [x] Integration tests pass in CI (17/17)


sdk/typescript/vitest.config.ts
```diff
@@ -7,6 +7,7 @@ export default defineConfig({
   test: {
     globals: true,
     environment: "node",
+    retry: process.env.CI ? 2 : 0,
```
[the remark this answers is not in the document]
x3c41a: Agreed, removed the retry config. Relying on increased timeouts only now.

sdk/typescript/src/preparer.ts
```diff
@@ -36,6 +36,7 @@ export class BulletinPreparer {
       defaultChunkSize: config?.defaultChunkSize ?? 1024 * 1024,
       createManifest: config?.createManifest ?? true,
       chunkingThreshold: config?.chunkingThreshold ?? 2 * 1024 * 1024,
+      txTimeout: config?.txTimeout ?? 120_000,
```
bkontur: @x3c41a @karolk91 `this.config` and all the default values are duplicated on two places, here and the second one: https://github.com/paritytech/polkadot-bulletin-chain/pull/389/files#diff-dc7dabfe111ec026bda4b1dd8194e9dd107bc9113bf990fbb68aa1af6134437bR564
```
this.config = {
      defaultChunkSize: config?.defaultChunkSize ?? 1024 * 1024, // 1 MiB
      createManifest: config?.createManifest ?? true,
      chunkingThreshold: config?.chunkingThreshold ?? 2 * 1024 * 1024, // 2 MiB
      txTimeout: config?.txTimeout ?? 120_000, // 2 minutes safety net
    }
```

We should deduplicate this default config and default values, to set them just once, or have just one place where we need to change any default values

sdk/typescript/src/async-client.ts
```diff
@@ -245,6 +248,11 @@ function mapPapiEventToProgress(
       }
     } else {
       progressCallback?.({ type: TxStatus.NoLongerInBlock, chunkIndex })
+      // Transaction dropped from best blocks and no longer valid
+      // (e.g. mortality expired, evicted from pool)
+      if (ev.isValid === false) {
+        result.isValid = false
```
bkontur: I don't know the code around , but from this diff `result.isValid` looks strange, also `PapiEventMappingResult` + `Result of mapping a single PAPI event to SDK progress events. - this is kind of confusing. 

If it means the mapping from PAPI event struct to SDK event struct - like some wrapper - then just the name `PapiEvent` should be enough. `MappingResult` and `Result of mapping` is pretty confusing at least to me. I think it should be either `Event` or `Result`. It cannot be both. `EventMappingResult` even `MappingResult` looks strange to me.

sdk/typescript/src/async-client.ts
```diff
@@ -192,6 +193,8 @@ interface PapiEventMappingResult {
     block: { hash: string; number: number }
     events?: RuntimeEvent[]
   }
+  /** Set to false when the transaction is no longer valid (e.g. mortality expired, dropped from pool) */
```
bkontur: shouldn't this doc go above?

sdk/typescript/src/async-client.ts
```diff
@@ -192,6 +193,8 @@ interface PapiEventMappingResult {
...
+  isValid?: boolean
```
bkontur: What is valid? Event or Result? I think at the end, it is about transaction validity.
Can the `PapiEventMappingResult` be `finish` and `isValid=false` at the same time from the API perspective?

sdk/typescript/src/preparer.ts
x3c41a: Done. Extracted `DEFAULT_CLIENT_CONFIG` in `types.ts` and replaced the inline defaults in all three constructors (`AsyncBulletinClient`, `MockBulletinClient`, `BulletinPreparer`) with `{ ...DEFAULT_CLIENT_CONFIG, ...config }`.

sdk/typescript/src/async-client.ts
x3c41a: Agreed, renamed to `MappedTxStatus`. It represents extracted transaction status from a PAPI event, not a generic mapping result.

sdk/typescript/src/async-client.ts
x3c41a: Moved the `isValid` description into the interface-level JSDoc block above. All three fields (`txHash`, `finish`, `isValid`) are now documented there.

sdk/typescript/src/async-client.ts
x3c41a: `isValid` refers to transaction validity (as reported by PAPI). They are mutually exclusive: `finish` requires the tx to be found in a block (`found=true`), while `isValid=false` only fires when PAPI reports `found=false, isValid=false` (mortality expired or evicted from pool). Added a comment in the interface doc to make this explicit.

sdk/typescript/src/async-client.ts
```diff
@@ -693,11 +711,20 @@ export class AsyncBulletinClient implements BulletinClientInterface {
...
           if (result.finish) finish(result.finish.block, result.finish.events)
+          // Transaction dropped from pool and no longer valid (mortality expired, evicted)
+          if (result.isValid === false) {
+            fail(
+              new BulletinError(
+                "Transaction is no longer valid (mortality expired or dropped from pool)",
+                ErrorCode.TIMEOUT,
+              ),
+            )
+          }
```
karolk91: As per my understanding, we cannot just throw error when `result.isValid === false` - its not a terminal signal, it only means that transaction is not valid at current best block - if there are forks it may become valid on another fork. Until block is finalized - `isValid` can change
```suggestion
```

x3c41a: Good catch. Removed the `isValid`-based error handling entirely. The configurable safety-net timeout (`txTimeout`) already covers stuck transactions.

sdk/typescript/src/types.ts
```diff
@@ -336,4 +337,20 @@ export interface ClientConfig {
...
+ * Default client configuration values.
+ *
+ * Used by AsyncBulletinClient, MockBulletinClient, and BulletinPreparer
+ * so that defaults are defined in one place.
+ */
+export const DEFAULT_CLIENT_CONFIG: Required<ClientConfig> = {
+  defaultChunkSize: 1024 * 1024, // 1 MiB
+  createManifest: true,
+  chunkingThreshold: 2 * 1024 * 1024, // 2 MiB
+  txTimeout: 120_000, // 2 minutes safety net
```
karolk91: I dont have deep understanding here to tell for sure, but what I've seen in practice (still its one of many possible scenarios) going with 2 minutes may not be correct. PAPI uses default mortality of txs (64 blocks?) which is roughly above 6 minutes given 6s blocks. We at least shouldn't terminate earlier than that..

I would question if we even need this `Safety-net timeout` at all. If you're willing to do more deep analysis then it would be better, but otherwise I'm ok with the default being 7 minutes or something like this

x3c41a: I see your point but realistically speaking, if 64 blocks delay happens on prod we'd extremely worried, won't we? I think an alert/page may fire after a few (<10) blocks. I don't mind increasing the delay to 64 blocks to be aligned with PAPI but IMO by doing that we'll just delay time to failure.

> still its one of many possible scenarios

just curious what those scenarios are? I just never seen them; not saying that they don't exist...

karolk91: its not about blockchain stopping producing blocks - its about transaction you just submitted at block X not included within X+64 blocks. Which happened to me a lot if for some reasons you used a nonce from the future (either by mistake or because you happened to be unlucky and got a nonce from a fork that was not finalized)

docs/book/src/typescript/error-handling.md
```diff
@@ -65,7 +65,7 @@ error.code === 'EMPTY_DATA'          // still works
 | Error Code | Retryable | Description | Recovery Hint |
 |---|---|---|---|
 | `TRANSACTION_FAILED` | Yes | On-chain transaction failed | Verify transaction parameters and account nonce |
-| `TIMEOUT` | Yes | Operation timed out | Increase timeout or retry |
+| `TIMEOUT` | Yes | Transaction not finalized within timeout window | Retry the transaction |
```
karolk91:
```suggestion
| `TIMEOUT` | Yes | Transaction not included in a block (best or finalized) within timeout window | Retry the transaction |
```

sdk/typescript/src/preparer.ts
```diff
@@ -32,11 +33,7 @@ export class BulletinPreparer {
   private config: Required<ClientConfig>
 
   constructor(config?: ClientConfig) {
-    this.config = {
-      defaultChunkSize: config?.defaultChunkSize ?? 1024 * 1024,
-      createManifest: config?.createManifest ?? true,
-      chunkingThreshold: config?.chunkingThreshold ?? 2 * 1024 * 1024,
-    }
+    this.config = { ...DEFAULT_CLIENT_CONFIG, ...config }
```
karolk91: refactor introduced a regression: passing { txTimeout: undefined } now produces undefined instead of falling back to the default. This would cause setTimeout(..., undefined) which fires immediately

sdk/typescript/src/types.ts
x3c41a: Ah, you're right. I double checked PAPI's code and`signSubmitAndWatch` actually errors with InvalidTxError once finalized blocks pass that mortality window (64 blocks). The existing error handler already catches that. So the timeout only fires if PAPI itself never reports back (connection drop, etc.)

Increasing to 7 mins sounds reasonable to prevent racing with PAPI's own timeout. The only thing that worries me is potentially degraded DX bc the CI already takes a longer while to finish

sdk/typescript/src/preparer.ts
x3c41a: Good catch. Added `resolveClientConfig()` that filters out undefined values before merging with defaults. `{ txTimeout: undefined }` now correctly falls back to the default.

sdk/typescript/src/async-client.ts
```diff
@@ -714,14 +715,14 @@ export class AsyncBulletinClient implements BulletinClientInterface {
         },
       })
 
-      // Timeout after 2 minutes
+      // Timeout per transaction (configurable, default 2 minutes)
       const timerId = setTimeout(() => {
```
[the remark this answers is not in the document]
voliva: PAPI doesn't have any timeout, as it's something that can always be done outside like in here.

However, connection drops should not cause PAPI to hang a tx indefinitely, unless the node completely goes down. PAPI reconnects and if after reconnecting it can link previously received blocks with new ones then continues tracking the transaction, otherwise it errors and stops tracking.

It could hang indefinitely though if the transaction is reported as valid through the `TaggedTransactionQueue_validate_transaction` API but the node for some reason rejects the transaction when broadcasting. But that's indicative of a bug.

tl;dr; PAPI should handle disconnects, but it's also fine to externally set up timeouts in CI. If they do timeout though, it's indicative of a potential issue.

sdk/typescript/src/types.ts
```diff
@@ -336,4 +337,36 @@ export interface ClientConfig {
...
+ * Default client configuration values.
+ *
+ * Used by AsyncBulletinClient, MockBulletinClient, and BulletinPreparer
+ * so that defaults are defined in one place.
+ */
+export const DEFAULT_CLIENT_CONFIG: Required<ClientConfig> = {
+  defaultChunkSize: 1024 * 1024, // 1 MiB
+  createManifest: true,
+  chunkingThreshold: 2 * 1024 * 1024, // 2 MiB
+  txTimeout: 420_000, // 7 minutes (above PAPI's 64-block mortality window)
```
voliva, line 356: FYI you can also configure mortality if you want a shorter or longer one:

```ts
tx.signSubmitAndWatch(bob, {
    mortality: {
      mortal: true,
      period: 16
    }
  })
```

(the period should be a power of 2... if it's not, PAPI will round it up to the next power of 2)

sdk/typescript/src/async-client.ts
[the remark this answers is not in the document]
x3c41a: Thanks for the clarification. Updated the comment to reflect that PAPI handles reconnects and mortality natively, and that this timeout is purely defensive - if it fires, it likely indicates a bug.

sdk/typescript/src/types.ts
x3c41a: Good to know, thanks. We're using the default mortality for now. If we need shorter timeouts in the future (e.g. for CI), configuring mortality directly would be the right approach.

karolk91 (approved):

PR #394, opened 2026-04-07
Implemet authorization extension without resetting the expiry

georgepisaltu:
Fixes #393 

This PR adds the option to extend an existing authorization without resetting its expiration. This is necessary for distributing anonymous bulletin authorizations in PoP. More details in the issue. The solution here is a new call to authorize an account, namely the `authorize_account_preserve_expiry` call. This call works like `authorize_account` but does not reset the expiration.


pallets/transaction-storage/src/lib.rs
```diff
@@ -569,23 +569,31 @@ pub mod pallet {
...
-		#[pallet::feeless_if(|origin: &OriginFor<T>, _who: &T::AccountId, _transactions: &u32, _bytes: &u64| -> bool {
+		#[pallet::feeless_if(|origin: &OriginFor<T>, _who: &T::AccountId, _transactions: &u32, _bytes: &u64, _refresh_expiry: &bool| -> bool {
 			T::Authorizer::try_origin(origin.clone()).is_ok()
 		})]
 		pub fn authorize_account(
 			origin: OriginFor<T>,
 			who: T::AccountId,
 			transactions: u32,
 			bytes: u64,
+			refresh_expiry: bool,
```
bkontur: @georgepisaltu we can go also with this breaking change, but we need to ensure that "People Paseo Stable" / "People Paseo Next" will be upgraded at the "same" time, so we don't block people

@karolk91 Is "People/Bulletin Paseo Review" using `authorize_account`? wdyt?

pallets/transaction-storage/src/lib.rs
```diff
@@ -618,11 +618,44 @@ pub mod pallet {
...
+		/// - `bytes`: The number of bytes that `who` may submit.
+		///
+		/// The origin for this call must be the pallet's `Authorizer`. Emits
+		/// [`AccountAuthorized`](Event::AccountAuthorized) when successful.
+		#[pallet::call_index(10)]
+		#[pallet::weight(T::WeightInfo::authorize_account())]
+		#[pallet::feeless_if(|origin: &OriginFor<T>, _who: &T::AccountId, _transactions: &u32, _bytes: &u64| -> bool {
+			T::Authorizer::try_origin(origin.clone()).is_ok()
+		})]
+		pub fn authorize_account_preserve_expiry(
```
karolk91: @georgepisaltu, could you please describe set of scenarios you would like authorizations to cover in general? In some illustrative way like

1. Bulletin block 1: PC sends XCM authorize Alice for 100 MB, 1tx -> expected state: Alice 100 MB, 1tx, expiry block: 1000
2. Bulletin block 5:  PC sends XCM authorize Alice for 50 MB, 2tx -> expected state: ?
3. Block 999 (1 before expiry): PC sends XCM authorize Alice
4. Block 1000: expected state all authorizations expired

Please consider edge cases like what happens when you want to add more storage at last block before expiry - will there be another XCM coming at the very next block to create some authorizations again?

My point is maybe we don't need the new extrinsic `authorize_account_preserve_expiry` - and just make regular `authorize_account` to never extend the expiry because we also have a separate extrinsic `refresh_account_authorization` to do the expiry refresh? If that works for all scenarios.

karolk91: Also not sure if you had a chance to read my design proposal about "slots" based authorizations where every `authorize_account` creates authorization with independent lifetime (up to some maximum slots per account) (and we then also get rid of  `refresh_account_authorization`) - if you haven't seen that let me know I will point you to a doc. I think `authorize_account_preserve_expiry` doesn't solve much problems when we can just use `refresh_account_authorization`

georgepisaltu: I didn't have a chance to read the proposal, but I generally agree with the idea, I think we should have slots for individual authorizations and then a counter per-user with the total allowance they have at any point in time.

That said, unless you can commit to delivering this in the next 2 weeks and have it integrated by the W3S, it doesn't work for us.

For the time being, we can deprecate the old call if it's not needed.

> Please consider edge cases like what happens when you want to add more storage at last block before expiry - will there be another XCM coming at the very next block to create some authorizations again?

This is fine, it's intended use, just like the call description says.

karolk91: @georgepisaltu, putting aside specific solutions for a moment. 

Could you describe scenarios to be supported here? Otherwise it's hard to reason about the solution. Right now I'm failing to see the flow here

If you authorize Alice at the last block before expiry of the current/existing authorizations using new extrinsic - Alice will have the sum of current + new authorizations but these will last only for 1 block. After 1 block - Alice will have 0 authorizations and then what should happen ?

georgepisaltu: I am saying this is the intended flow, if Alice is claiming an authorization on the People chain 1 block before her previous authorization on the same account expires, we want it to expire in 1 block. The earliest Alice can claim an authorization in one period is at the first block of the period, and the authorization lasts for a whole period (2 weeks). Alice can reauthorize the account when the next period starts (in 1 block in your example). Again this is not only fine, it is required in order to prevent spam in PoP-issued authorizations.

You can find a description of why we are doing this in the issue linked in this PR description, #393 . This will be used in https://github.com/paritytech/individuality/pull/785 , which was recently updated with https://github.com/franciscoaguirre/individuality/pull/1 to support anonymous PoP authorizations.

After this PR, as a follow up we will need to introduce a grace period before allowing data to be removed (in case someone needs to renew the authorization to some anonymous account to keep the data). In the long term, we should do the slot-based solution because what we have now is worse than that, but there is little time left until we have to release.

Unfortunately discussions about this solution are now lost in chats, but I hope there is sufficient context to move forward with this.

georgepisaltu: One more thing I wanted to add: if you don't need the authorization call to refresh the authorization in any scenario or use case, then I am happy to trim this PR and just change that one bit in the original extrinsic and move forward with that while we wait for the slot based authorizations. The purpose of the change in this PR is to avoid a breaking change, I don't know how the bulletin chain is used outside of our PoP use cases.

karolk91: > One more thing I wanted to add: if you don't need the authorization call to refresh the authorization in any scenario or use case, then I am happy to trim this PR and just change that one bit in the original extrinsic and move forward with that while we wait for the slot based authorizations. The purpose of the change in this PR is to avoid a breaking change, I don't know how the bulletin chain is used outside of our PoP use cases.

I think that you are aware of almost all authorize scenarios out there since in production network these will only come from People Chain / Asset Hub, right? @bkontur, @franciscoaguirre wdyt? Should we modify existing authorize_account to never extend the expiration ? Could possibly someone rely on this - and maybe a better question if they do, should they rely on this (IMHO, no) ?

karolk91: > if Alice is claiming an authorization on the People chain 1 block before her previous authorization on the same account expires, we want it to expire in 1 block.

Is this triggered by the user itself ? Like a button in the application user needs to click ? I would think this is somehow automatic / behind the scenes as it doesn't seem a good UX for an user to keep renewing his authorizations

franciscoaguirre: I'm fine with just changing the behavior of `authorize_account` to make it not refresh expiration and point current users to `refresh_account_authorization`. My main concern was changing the public API and having coordination issues

franciscoaguirre (approved):

pallets/transaction-storage/src/tests.rs
```diff
@@ -1297,3 +1297,65 @@ fn authorize_storage_extension_passes_through_non_storage_calls() {
...
+		run_to_block(10, || None);
+		assert_ok!(TransactionStorage::validate_signed(&who, &call));
+
+		// Expires at block 11 (original expiry), NOT 15.
+		run_to_block(11, || None);
+		assert_eq!(
+			TransactionStorage::account_authorization_extent(who),
+			AuthorizationExtent { transactions: 0, bytes: 0 },
+		);
+		assert_noop!(TransactionStorage::validate_signed(&who, &call), InvalidTransaction::Payment);
```
karolk91, line 1328: lets also consider a test case that `authorize_account` still works when authorization is already expired - should create a new authorization with fresh expiry. Or maybe just extend this one with more steps

karolk91 (approved):

bkontur (approved):

PR #395, opened 2026-04-08
Add renew support for rootCID to console-ui 

antkve:
Addresses #359


console-ui/src/lib/cid-lookup.ts
```diff
@@ -0,0 +1,183 @@
...
+  return cid.code === CidCodec.DagPb;
+}
+
+/** 32-byte multihash digest as 0x-hex — the on-chain `content_hash` shape. */
+export function contentHashHex(cid: CID): string {
+  return Binary.toHex(cid.multihash.digest);
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type Api = any;
```
franciscoaguirre: We can use typed apis here, can't we?

console-ui/src/lib/cid-lookup.ts
```diff
@@ -0,0 +1,183 @@
...
+
+interface TxInfo {
+  content_hash: HexString;
+  size: number;
+}
+
+async function locateContentHashes(
```
franciscoaguirre, line 35: Why do we need this and it's not enough with `TransactionByContentHash`?

console-ui/src/pages/Renew/Renew.tsx
```diff
@@ -185,6 +235,175 @@ export function Renew() {
...
+    setCidInput(value);
+    setIsCidValid(isValid);
+    setParsedCid(cid);
+    // Clear previous resolution when CID changes
+    setResolutions([]);
+    setTotalSize(0);
+    setCheckedCids(new Set());
+    setResolveError(null);
+    setBatchError(null);
+    setBatchResults([]);
```
franciscoaguirre, line 258: Maybe this should be one struct if it's updated at the same time

franciscoaguirre (approved):

PR #403, opened 2026-04-10
Github actions in depandabot

RafalMirowski1:
- closes https://github.com/paritytech/polkadot-bulletin-chain/issues/391
- pins actions to SHA
- adds github actions to dependabot


.github/workflows/check-console-ui.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     name: Check Console UI
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
```
karolk91:
```suggestion
      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
```

seems like we have been inconsitent with versions usage - like above
could track and unify all such occurences ?

RafalMirowski1: unified the versions, one discrepency left is `uses: dtolnay/rust-toolchain@631a55b12751854ce901bb631d5902ceb48146f7 # stable` in `release.yml` vs `uses: dtolnay/rust-toolchain@efa25f7f19611383d5b0ccf2d1c8914531636bf9 # master` with explicit `toolchain: "${{ env.RUST_STABLE_VERSION }}"` in `check-runtime-migration.yml` but idk if thats intentional

bkontur (approved):

PR #404, opened 2026-04-13
Added check-runtime to CI

antkve:
---

_Part of #632 — cross-repo roadmap tracking._


runtimes/bulletin-polkadot/src/lib.rs
```diff
@@ -558,35 +557,66 @@ where
...
-		BridgePolkadotGrandpa: pallet_bridge_grandpa = 51,
-		BridgePolkadotParachains: pallet_bridge_parachains = 52,
-		BridgePolkadotMessages: pallet_bridge_messages = 53,
-
-		// Local Root
-		Sudo: pallet_sudo = 61,
-		Proxy: pallet_proxy = 62,
-	}
-);
+#[frame_support::runtime(legacy_ordering)]
```
bkontur, line 560: @antkve we don't need to modify bulletin-polkadot solochain runtime, it will be remove soon

.github/workflows/check-runtime-migration.yml
```diff
@@ -134,6 +134,74 @@ jobs:
           key: try-runtime-snapshot-${{ matrix.runtime.name }}-${{ steps.get-date.outputs.today }}
 
 
+  check-runtimes:
```
bkontur, line 147: @antkve please, add here some specific/short desc what `@polkadot-api/check-runtime@latest` is really checking

.github/workflows/check-runtime-migration.yml
```diff
@@ -134,6 +141,77 @@ jobs:
           key: try-runtime-snapshot-${{ matrix.runtime.name }}-${{ steps.get-date.outputs.today }}
 
 
+  # Runs @polkadot-api/check-runtime against live RPCs to detect common runtime
+  # upgrade issues: missing or outdated metadata versions, absent runtime APIs,
+  # incorrect CheckMetadataHash configuration, and dev APIs in production builds.
+  check-runtimes:
+    needs: [runtime-matrix]
+    runs-on: ubuntu-latest
+    if: github.event_name != 'schedule' && github.event_name != 'workflow_dispatch' && needs.runtime-matrix.outputs.check-runtime != '[]'
```
bkontur: @antkve can we just use that bool directly?

```suggestion
    if: github.event_name != 'schedule' && github.event_name != 'workflow_dispatch' && needs.runtime-matrix.outputs.runtime.check-runtime
```

I would remove:
```
# Filter runtimes with check_runtime enabled (metadata hash matches the live chain)
          CHECK_RUNTIME_TASKS=$(jq -c '[.[] | select(.check_runtime == true)]' scripts/runtimes-matrix.json)
          echo --- check-runtime tasks ---
          echo $CHECK_RUNTIME_TASKS
          echo "check-runtime=$CHECK_RUNTIME_TASKS" >> $GITHUB_OUTPUT
```

antkve: @bkontur Yes have changed to use the bool directly. But if I remove that filter, the checks fail at the moment because paseo builds with the wrong metadata (it builds with enable_metadata_hash("WND", 12)). Have made a PR to build Paseo with the correct hash using build flags, changes here: https://github.com/paritytech/polkadot-bulletin-chain/pull/411/changes/e783decd9b6b6d955a4df1661fe8c0c884016275

bkontur:
@antkve can you please rebase?

PR #405, opened 2026-04-14
Make authorizer list configurable

### LINKED ISSUE #63: Make `Authorizer` configurable
Actually, only the `Root` can authorize, but we want to be more flexible (for testing and handling more use-cases).

### Solution
- [ ] Introduce storage item `pub type AllowedAuthorizers` to the `transaction-store` pallet
- [ ] Add `GenesisConfig` with `allowed_authorizers`
- [ ] Add very simple extrinsics `add_authorizers` / `remove_authorizers` for `ensure_root`
- [ ] Add `EnsureOrigin` impl `EnsureAllowedAuthorizers` which checks for `AllowedAuthorizers`
```diff
- type Authorizer = EnsureRoot<Self::AccountId>;
+ type Authorizer = EitherOf<
+        EnsureRoot<Self::AccountId>
+        EnsureAllowedAuthorizers,
+ >;
```
- [ ] Add pallet/runtime unit-tests
- [ ] Bootstrap `AllowedAuthorizers` with `//Alice` / `//Bob` with genesis preset for `LOCAL_TESTNET_RUNTIME_PRESET`

ilchu:
Closes https://github.com/paritytech/polkadot-bulletin-chain/issues/63

### Summary
This PR replaces a hardcoded list of authorizers with a dynamic structure that allows their addition and removal in runtime.

### Changes
* A new storage item in `transaction_store` pallet and two extrinsics, `add_authorizer` and `remove_authorizer`
* A custom `ManagerOrigin` for managing the authorizer list
* Two benchmarks for the new calls and regenerated extrinsic weights (not sure I should have done it locally)
* Genesis additional field and population logic
* A new migration to preserve the extra authorizer introduced in #381
* Tests in the pallet and both runtimes


pallets/transaction-storage/src/lib.rs
```diff
@@ -452,24 +485,6 @@ pub mod pallet {
...
-		/// Index and store data off chain with an explicit CID configuration.
-		///
-		/// Behaves identically to [`store`](Self::store), but the CID configuration
-		/// (codec and hashing algorithm) is passed directly as a parameter.
-		///
-		/// Emits [`Stored`](Event::Stored) when successful.
-		#[pallet::call_index(9)]
-		#[pallet::weight(T::WeightInfo::store(data.len() as u32))]
-		#[pallet::feeless_if(|_origin: &OriginFor<T>, _cid: &CidConfig, _data: &Vec<u8>| -> bool { true })]
-		pub fn store_with_cid_config(
```
franciscoaguirre, line 445: Why was this moved?

pallets/transaction-storage/src/lib.rs
```diff
@@ -706,6 +721,65 @@ pub mod pallet {
...
+
+		/// Add an account to the set of allowed authorizers. Allowed authorizers can call
+		/// [`authorize_account`](Self::authorize_account) and
+		/// [`authorize_preimage`](Self::authorize_preimage) to grant storage access.
+		///
+		/// If the account is already an allowed authorizer, this is a no-op.
+		///
+		/// Parameters:
+		///
+		/// - `who`: The account to add as an allowed authorizer.
```
franciscoaguirre: We need to also add a budget that this authorizer can authorize for.

An override of the authorization period would also be useful, even though it might require a lot of changes.

pallets/transaction-storage/src/lib.rs
```diff
@@ -706,6 +721,65 @@ pub mod pallet {
...
+		///
+		/// The origin for this call must satisfy `ManagerOrigin`. Emits
+		/// [`AuthorizerAdded`](Event::AuthorizerAdded) when successful.
+		#[pallet::call_index(10)]
+		#[pallet::weight(T::WeightInfo::add_authorizer())]
+		pub fn add_authorizer(origin: OriginFor<T>, who: T::AccountId) -> DispatchResult {
```
franciscoaguirre: We could set Locations as authorizers. It's more inline with what we currently use. However, we can use the sovereign accounts for locations instead

pallets/transaction-storage/src/lib.rs
```diff
@@ -706,6 +721,65 @@ pub mod pallet {
...
+		///
+		/// Parameters:
+		///
+		/// - `who`: The account to remove from the allowed authorizers.
+		///
+		/// The origin for this call must satisfy `ManagerOrigin`. Emits
+		/// [`AuthorizerRemoved`](Event::AuthorizerRemoved) when successful.
+		#[pallet::call_index(11)]
+		#[pallet::weight(T::WeightInfo::remove_authorizer())]
+		pub fn remove_authorizer(origin: OriginFor<T>, who: T::AccountId) -> DispatchResult {
```
franciscoaguirre, line 877: When we add the "budget" for these authorizers, we could also allow anyone to remove authorizers who've exhausted all their budget

pallets/transaction-storage/src/lib.rs
```diff
@@ -778,6 +861,8 @@ pub mod pallet {
 		pub byte_fee: BalanceOf<T>,
 		pub entry_fee: BalanceOf<T>,
 		pub retention_period: BlockNumberFor<T>,
+		/// Initial accounts that are allowed to issue authorizations.
```
franciscoaguirre:
```suggestion
		/// Initial additional accounts that are allowed to issue authorizations.
```

Because this is one of many Authorizers config items

pallets/transaction-storage/src/lib.rs
ilchu: Just a convenience choice to sort by call index as I was confused at first, having missed that index 9 call already exists.

pallets/transaction-storage/src/lib.rs
ilchu: I'll try to add it like the other `remove_expired_..` calls as a separate extrinsic. And I suppose we still want to keep the sudo way of removing them as well.

pallets/transaction-storage/src/lib.rs
ilchu: From what I understand, the location -> account ID is a pretty straightforward and automatic conversion, whereas the other way is not so much, so I agree we can leave the current authorizer storage map keyed as is.

pallets/transaction-storage/src/lib.rs
ilchu: I was thinking through this and one question that rose was having sane defaults. 
I.e. if we enforce budgeting when adding authorizers, what would the budgets be for the authorizers present since genesis? And will giving them virtually unlimited budget be acceptable?

bkontur:
/cmd bench --runtime bulletin-westend --pallet pallet_bulletin_transaction_storage

bkontur:
/cmd bench --runtime bulletin-paseo --pallet pallet_bulletin_transaction_storage

bkontur (approved):

PR #409, opened 2026-04-14
Rust sdk wait for block inclusion

RafalMirowski1:
- Add WaitFor enum (InBlock / Finalized) to control when transaction submission methods return, replacing the unused wait_for_finalization: bool field    
- Resolve nonces from the best block before each submission, so sequential InBlock transactions work without waiting for finalization                                                                                                                                              
- Unify duplicated tx-status streaming loops into a single submit_and_watch method used by all 9 public transaction methods                               
- Update docs and examples


bkontur:
@RafalMirowski1 still draft?

sdk/rust/src/transaction.rs
```diff
@@ -45,6 +52,165 @@ impl TransactionClient {
...
+			.await
+			.map_err(|e| make_error(format!("{e:?}")))
+	}
+
+	/// Submit a transaction, stream status events, and return when the target
+	/// confirmation level is reached.
+	///
+	/// This is the single submission loop used by all public methods. It:
+	/// - Streams all `TxStatus` events, firing the optional progress callback
+	/// - Breaks on `InBestBlock` or `InFinalizedBlock` based on `wait_for`
```
ilchu:
```suggestion
	/// - Breaks on `InBestBlock` or `InFinalizedBlock` depending on the value of `wait_for`
```

sdk/rust/src/transaction.rs
```diff
@@ -45,6 +52,165 @@ impl TransactionClient {
...
+							if let Some(ref cb) = progress_callback {
+								cb(ProgressEvent::tx_validated());
+							},
+						TxStatus::Broadcasted =>
+							if let Some(ref cb) = progress_callback {
+								cb(ProgressEvent::tx_broadcasted());
+							},
+						TxStatus::InBestBlock(in_block) => {
+							let block_hash = format!("{:?}", in_block.block_hash());
+							let extrinsic_hash = format!("{:?}", in_block.extrinsic_hash());
```
ilchu: It's a nit, but calling `to_string()` here and in `InFinalizedBlock` would arguably be cleaner.

sdk/rust/src/types.rs
```diff
@@ -280,23 +280,39 @@ pub struct RenewalResult {
...
 /// Options for storing data.
 #[derive(Debug, Clone, Encode, Decode, TypeInfo)]
 pub struct StoreOptions {
 	/// CID codec to use (default: raw).
 	pub cid_codec: CidCodec,
 	/// Hashing algorithm to use (default: blake2b-256).
 	pub hash_algorithm: HashingAlgorithm,
-	/// Whether to wait for finalization (default: false).
-	pub wait_for_finalization: bool,
+	/// What to wait for before returning (default: InBlock).
```
ilchu:
```suggestion
	/// What to wait for before returning (default: best block inclusion).
```
or something along these lines? Just to make it a bit more descriptive.

sdk/rust/src/transaction.rs
```diff
@@ -45,6 +52,165 @@ impl TransactionClient {
 		&self.api
 	}
 
+	/// Submit a transaction with the nonce resolved from the best block.
+	///
+	/// Subxt's default nonce resolution queries the finalized block, which
+	/// lags behind for sequential `WaitFor::InBlock` transactions. Instead,
+	/// we query at the current best block
```
ilchu:
```suggestion
	/// Submit a transaction with the nonce resolved from the best block.
	///
	/// Subxt's default nonce resolution queries the finalized block which
	/// lags behind actual block inclusion for sequential `WaitFor::InBlock` transactions. So we manually override that by choosing to
	/// query at the current best block instead.
```

sdk/rust/src/transaction.rs
```diff
@@ -127,6 +298,10 @@ impl TransactionClient {
 	/// Returns an error immediately if authorization is missing, expired,
 	/// or insufficient (avoiding a wasted transaction submission).
 	///
+	/// The `wait_for` parameter controls when the method returns:
+	/// - `WaitFor::InBlock` - Returns after transaction is included in a best block (faster)
+	/// - `WaitFor::Finalized` - Returns after transaction is finalized (safer)
+	///
```
ilchu: I'm a bit confounded as to why we would include such a detailed description of the `WaitFor` struct here when it has its own docstring. And if we do, why not add it to other methods receiving it as a parameter as well.

ilchu (changes requested):
Pretty much LGTM, just petty things around doc clarity mostly.

ilchu (approved):

docs/book/src/README.md
```diff
@@ -77,7 +77,7 @@ use bulletin_sdk_rust::prelude::*;
 
 let client = TransactionClient::new("wss://paseo-bulletin-rpc.polkadot.io").await?;
 let data = b"Hello, Bulletin!".to_vec();
-let receipt = client.store(data, &signer).await?;
+let receipt = client.store(data, &signer, WaitFor::InBlock).await?;
```
bkontur, line 80: @RafalMirowski1 what are the examples for TS SDK useing? InBlock or Finalized? Should be aligned.

bkontur (approved):
@RafalMirowski1 please fix fmt job

PR #410, opened 2026-04-14
Build docker image for stress test tool

sandreim:
---

_Part of #627 — cross-repo roadmap tracking._


.github/workflows/build-stress-test-image.yml
```diff
@@ -0,0 +1,82 @@
...
+
+env:
+  IMAGE_NAME: docker.io/paritypr/bulletin-stress-test
+
+jobs:
+  check:
+    name: Build & verify binary
+    runs-on: ubuntu-latest
+    container:
+      image: paritytech/ci-unified:bullseye-1.88.0-2025-06-27-v202507112050
```
bkontur: Can we use the image from `env` file https://github.com/paritytech/polkadot-bulletin-chain/blob/main/.github/env#L1 ?

.dockerignore
```diff
@@ -0,0 +1,10 @@
+target/
+.git/
+node_modules/
+docs/
+examples/
+zombienet/
```
bkontur, line 6: If you need just image for stress tool, I would exclude everyhting else

```suggestion
zombienet/
sdk/
runtimes/
scripts/
templates/
pallets/
node/
console-ui/
```

PR #411, opened 2026-04-15
Add build flags to build Paseo with correct metadata

antkve:


antkve:
Hey @bkontur @karolk91 @franciscoaguirre ,

I think this metadata build flag PR should be expanded into one that feature-gates a lot more than just the metadata. I got Claude to run some XCM tests and it seems a bunch of stuff on Bulletin Paseo is still configured for westend:

## Results

**Test 3 — ExistentialDeposit:** Identical `1_000_000_000` plancks on both chains.
- On Westend (12 decimals): 0.001 WND — reasonable
- On Paseo (10 decimals): **0.1 PAS — 100× too high** ← Westend constants confirmed live on Paseo

**Test 5 — The smoking gun (`LocationToAccountApi.convert_location`):**

On **live Paseo bulletin**, probing `AccountId32Aliases` with three network identifiers:

| Network passed                                   | Result                  |
| ------------------------------------------------ | ----------------------- |
| `ByGenesis(WESTEND_GENESIS_HASH)`                | ✅ `Ok(...)` — **accepted as local** |
| `ByGenesis(PASEO_GENESIS_HASH)`                  | ❌ `Err(Unsupported)` — rejected |
| `None`                                           | ✅ `Ok(...)` — accepted (default) |

**The live Paseo bulletin chain treats Westend's genesis hash as its local relay network and rejects its own Paseo genesis hash.** Westend bulletin behaves identically — it's the same WASM.

Full report saved to `/tmp/paseo-test/REPORT.md`.

## Implications

1. **Outbound XCM corruption:** Any XCM sent from Paseo bulletin (e.g. to Asset Hub Paseo) will carry `UniversalOrigin { network: ByGenesis(WESTEND_GENESIS_HASH) }`. The receiver's `AccountId32Aliases` won't recognize it as a sibling on the Paseo consensus system, so sovereign-account derivation will land on a completely different address than intended.
2. **Inbound XCM corruption:** Messages from genuine Paseo-network origins carrying `AccountId32 { network: ByGenesis(PASEO_GENESIS_HASH) }` are returned `Unsupported` by the bulletin chain's location-to-account resolver — they'll fall through to `HashedDescription`, producing an unrelated account.
3. **100× ED miscalibration** on Paseo: a user funding an account with 0.099 PAS will fail to create it; the Westend economics (where 0.001 WND ≈ $0.00001) translate to 0.1 PAS being required as dust, which is significant.

## Why no test caught this

- The XCM emulator integration tests in `runtimes/bulletin-westend/integration-tests/` are hardcoded to `westend_emulated_chain::Westend` and `asset_hub_westend_emulated_chain::AssetHubWestend`. There's no Paseo emulator.
- Paseo's `integration_tests: true` runs only storage/SDK tests that never construct XCM, and use the chain spec decimals (correct) rather than runtime currency constants (incorrect).
- `check-runtime-migration` validates upgrade semantics on a state snapshot but can't detect a hardcoded relay network ID.

## Recommended fix

Feature-flag or parameterize via a new module (e.g. `network_config::westend` vs `network_config::paseo`) gating:
- `RelayNetwork` hash
- Currency/fee/ED constants (Paseo mirrors Polkadot's 10-decimal economics, so reuse the polkadot/rococo constants or add a paseo module)
- System parachain IDs (`ASSET_HUB_ID`, `COLLECTIVES_ID`)
- Governance / People chain locations

The existing `metadata-hash-paseo` Cargo feature is the natural switch to extend; turn it into a proper `paseo` feature that gates all of these.

Additionally, a Paseo XCM emulator test (new integration-tests crate mirroring the Westend one) would have caught this at PR-time — that's the long-term structural fix.


Thoughts?

PR #417, opened 2026-04-15
Enable elastic scaling for testnet runtime

sandreim:
---

_Part of #623 — cross-repo roadmap tracking._


bkontur:
@sandreim probably something is missing, because it does not produce blocks for actual setup:  https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/24466738757?pr=417 and at the button there are `failed-zombienet-logs-bulletin-westend`, I remember at least locally this worked for 3 cores and 2s blocks for me: https://github.com/paritytech/polkadot-bulletin-chain/pull/231


@karolk91 if we want/need to merge this to the bulletin-westend, then please let's create the exact copy as `bulletin-paseo`, because we don't want to deploy ES to Paseo nets, let's keep two separate copies just in case, we don't deploy anything unexpected to the Paseo Review or Next

sandreim:
> @sandreim probably something is missing, because it does not produce blocks for actual setup: https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/24466738757?pr=417 and at the button there are `failed-zombienet-logs-bulletin-westend`, I remember at least locally this worked for 3 cores and 2s blocks for me: #231
> 
> @karolk91 if we want/need to merge this to the bulletin-westend, then please let's create the exact copy as `bulletin-paseo`, because we don't want to deploy ES to Paseo nets, let's keep two separate copies just in case, we don't deploy anything unexpected to the Paseo Review or Next

Yeah, node needs to run with `--authoring=slot-based`

bkontur:
We have dedicated Paseo runtime, so for westend we can even merge this probably

sandreim:
@karolk91  You can keep the block processing velocity to 3, chain will run 6s blocks if it only has 1 core.

PR #422, opened 2026-04-17
Stress test fixes

sandreim:


bkontur (approved):
@sandreim can you please fix conflicts, I had to push fmt for stress-test module, because the fmt CI fails for every PR now. Alternatively, we can exclude stress-test from fmt check

sandreim:
> @sandreim can you please fix conflicts, I had to push fmt for stress-test module, because the fmt CI fails for every PR now. Alternatively, we can exclude stress-test from fmt check

Looks like I pushed them by mistake here: https://github.com/paritytech/polkadot-bulletin-chain/pull/417/changes 🤦🏼

PR #427, opened 2026-04-20
Fix flaky integration tests: use production txTimeout

x3c41a:
We tried increasing txTimeout from 60s to 120s in CI. Both kept getting exceeded by slow zombienet finalization on loaded runners.

The production default of 420s exists for a reason: it sits above PAPI's 64-block mortality window and lets PAPI handle retries as designed. Overriding it lower in tests creates **a failure mode that real users never hit.**

Drops the override entirely and bumps per-test timeouts on chunked tests from 180s to 600s to match.

### DIFF
--- a/sdk/typescript/test/integration/client.test.ts
+++ b/sdk/typescript/test/integration/client.test.ts
@@ -53,13 +53,12 @@ describe("AsyncBulletinClient Integration Tests", { timeout: 120_000 }, () => {
     )
     aliceAddress = ss58Address(aliceKeyPair.publicKey, 42)
 
-    // Create client directly with api, signer, and submit function
-    // Per-tx timeout for CI zombienet nodes. 60s was too aggressive —
-    // finalization regularly takes >60s under CI load, causing flaky
-    // "Transaction timed out" failures on chunked store tests.
-    client = new AsyncBulletinClient(api, signer, papiClient.submit, {
-      txTimeout: 120_000,
-    })
+    // Use the production default txTimeout (420s). Previous attempts to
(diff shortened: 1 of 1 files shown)


sdk/typescript/test/integration/client.test.ts
```diff
@@ -159,7 +158,7 @@ describe("AsyncBulletinClient Integration Tests", { timeout: 120_000 }, () => {
     })
 
     it("should fire progress events in correct order during chunked upload", {
-      timeout: 180_000,
+      timeout: 600_000,
```
bkontur: 10 min? I would go with max 2 * 180_000

bkontur: or why not 420s or 480s?

x3c41a:
Good call @bkontur, 600s was too generous. Dropped to 480s: just above the 420s single-tx timeout, covers normal multi-tx execution plus one slow tx.

sdk/typescript/test/integration/client.test.ts
x3c41a: too generous, I agree. 

Changed to 480 (prefer extra safety net)

bkontur (approved):

bkontur:
@RafalMirowski1 @karolk91 please, please also check before merging

RafalMirowski1 (approved):

karolk91:
> @RafalMirowski1 @karolk91 please, please also check before merging

@x3c41a is "loaded runner" really an explanation to the issues here ? I find it very suspicious that it takes such a long time - maybe we are doing something wrong on the SDK side

PR #429, opened 2026-04-20
Fix flaky CI: handle PAPI Observable completion in tx subscription

x3c41a:
Supersedes #427.

The recurring "Transaction timed out" CI failures weren't caused by slow runners. The actual problem: PAPI's `signSubmitAndWatch` Observable can **complete without emitting** a finalized/in_block event, and we had no `complete` handler.

When a zombienet node restarts under CI load (the "heartbeat timeout" messages in failed runs), `txBestBlocksState` fires with `found: false`. PAPI's internal `continueWith()` operator maps that to `rxjs.EMPTY`, which completes the Observable immediately. Our Promise then hangs silently until the defensive timeout fires minutes later.

This adds a `complete` handler that rejects immediately with `TRANSACTION_FAILED` so the error surfaces fast and callers can retry.


bkontur:
@x3c41a Something like this? https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/24710314616/job/72273663389?pr=432

sdk/typescript/src/async-client.ts
```diff
@@ -734,6 +735,25 @@ export class AsyncBulletinClient implements BulletinClientInterface {
...
+          // PAPI can complete the Observable without a finalized/in_block
+          // event (e.g. txBestBlocksState fires with found:false after a
+          // reorg or node restart, causing the internal continueWith() to
+          // map to rxjs.EMPTY which completes immediately). Without this
+          // handler the Promise hangs until the defensive timeout fires.
+          if (!resolved) {
+            resolved = true
+            cleanup()
+            reject(
+              new BulletinError(
```
bkontur, line 754: @x3c41a what about just something like this?
```
progressCallback({
                type: isDropped ? TxStatus.Dropped : TxStatus.Invalid,
                error: errorMsg,
                chunkIndex,
              })
```

what about the case: `if (resolved) {` should we do `progressCallback` or some console.log or anything?

x3c41a: If it comes to progressCallback: I fixed in latest commit. Now it always returns `TxStatus.Dropped` here since we know the subscription ended without delivering the result. 

On `if (resolved)`: That's the normal success path, tx already went through. Nothing to report there.

x3c41a:
> @x3c41a Something like this? https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/24710314616/job/72273663389?pr=432

Yes, same issue. That branch doesn't have the complete handler so it hangs until the timeout.

bkontur (approved):

bkontur:
hmm, looks like this does not help, or there is another issue: https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/24765323268/job/72458357994?pr=368#step:22:146

@x3c41a please, check this

PR #435, opened 2026-04-21
SDK: multi-account pipeline submission

x3c41a:
**Multi-account does not beat single-account on Versi**

1. For **500 x 128KB** throughput flat at ~5 tx/s; per-user finalization grows 37s → 72s as accounts go 1 → 4.
2. Fro **1000 * 100KB** single-account already hits ~27 tx/s; multi-account does not help.

### 1000 × 10 KB (10 MB total, headroom under the 512 tx/block cap)

1000 × 10 KB

Single-account already reaches ~27 tx/s here, well below the 512 tx/block hard cap, which suggests pool / RPC / block-production throughput is the limit rather than nonce contention. Multi-account does not push past it.

### 500 × 128 KB (62.5 MB, block fully saturated at ~73 tx/block)

500 × 128 KB


x3c41a:
Closing. Multi-account does not beat single-account on Versi

PR #442, opened 2026-04-22
Fix flaky CI: remove txTimeout override in integration tests

x3c41a:
The 120s txTimeout override in integration tests was too tight. Latest failure at 126s, while the JS examples use 180s with the same PAPI `signSubmitAndWatch` and never flake.

Rather than bump to another arbitrary value, this removes the override entirely so tests use the production default (420s). Same timeout real users get.

Follows up on #429.

---

_Part of #632 — cross-repo roadmap tracking._

### DIFF
--- a/sdk/typescript/test/integration/client.test.ts
+++ b/sdk/typescript/test/integration/client.test.ts
@@ -53,13 +53,10 @@ describe("AsyncBulletinClient Integration Tests", { timeout: 120_000 }, () => {
     )
     aliceAddress = ss58Address(aliceKeyPair.publicKey, 42)
 
-    // Create client directly with api, signer, and submit function
-    // Per-tx timeout for CI zombienet nodes. 60s was too aggressive —
-    // finalization regularly takes >60s under CI load, causing flaky
-    // "Transaction timed out" failures on chunked store tests.
-    client = new AsyncBulletinClient(api, signer, papiClient.submit, {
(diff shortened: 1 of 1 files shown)


sdk/typescript/test/integration/client.test.ts
```diff
@@ -53,13 +53,10 @@ describe("AsyncBulletinClient Integration Tests", { timeout: 120_000 }, () => {
     )
     aliceAddress = ss58Address(aliceKeyPair.publicKey, 42)
 
-    // Create client directly with api, signer, and submit function
-    // Per-tx timeout for CI zombienet nodes. 60s was too aggressive —
-    // finalization regularly takes >60s under CI load, causing flaky
-    // "Transaction timed out" failures on chunked store tests.
-    client = new AsyncBulletinClient(api, signer, papiClient.submit, {
-      txTimeout: 120_000,
```
bkontur, line 61: I would keep this explicitely: `txTimeout: 180_000,`

bkontur (changes requested):
please https://github.com/paritytech/polkadot-bulletin-chain/pull/442/files#r3123078709

bkontur:
here it was pretty fast: https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/24775310863/job/72492234529?pr=447#step:22:141

PR #447, opened 2026-04-22
More bumps

bkontur:


rosarp (approved):

x3c41a (approved):

.github/env
```diff
@@ -1,16 +1,16 @@
 CI_IMAGE=paritytech/ci-unified:bullseye-1.88.0-2025-06-27-v202507112050
 RUST_STABLE_VERSION=1.88.0
 NODE_VERSION=22
-POLKADOT_SDK_VERSION=polkadot-stable2512
-POLKADOT_SDK_VERSION_WITH_BULLETIN=unstable-bulletin-support-v1
+POLKADOT_SDK_VERSION=polkadot-stable2603
+POLKADOT_SDK_VERSION_WITH_BULLETIN=polkadot-stable2603
```
karolk91: Maybe remove it if its now the same? will also require to replace `POLKADOT_SDK_VERSION_WITH_BULLETIN` in other places with `POLKADOT_SDK_VERSION`
```suggestion
```

bkontur: good point, but I would remove this once we get HOP and auto-renewals in, just in case, we don't need this special handling :)

karolk91:
We likely don't have any HOP scenarios in the repo? As far as I know HOP is not part of `stable2603`

franciscoaguirre:
> We likely don't have any HOP scenarios in the repo? As far as I know HOP is not part of `stable2603`

Not part of stable2603 no

franciscoaguirre (approved):

PR #448, opened 2026-04-22
Adds `AllowanceBasedPriority` extension

bkchr:
This extension gives `store` transactions a proportional priority boost based on the already consumed allowance.

## TODO

- [x] revert auth call and just ignore tx count or clean all the SDK and examples (breaking change)
- [x] fix migration for Paseo
- [ ] upgrade Paseo
- [x] Align with https://github.com/paritytech/individuality/pull/785 and https://github.com/paritytech/triangle-js-sdks/pull/129


pallets/transaction-storage/src/lib.rs
```diff
@@ -997,40 +997,30 @@ pub mod pallet {
...
 					} else {
-						// An unexpired authorization already exists. Extend it.
 						match scope {
 							AuthorizationScope::Account(_) => {
-								// Add
-								authorization.extent.transactions =
-									authorization.extent.transactions.saturating_add(transactions);
-								authorization.extent.bytes =
-									authorization.extent.bytes.saturating_add(bytes);
+								authorization.extent.bytes_allowance = bytes_allowance;
```
bkontur: @bkchr this one I am not sure yet, before we increased the allowance, but now we just reset. I think this is definitely related to the https://github.com/paritytech/individuality/pull/785

```suggestion
								authorization.extent.bytes_allowance = authorization                                                                                                                                                
									.extent                                                                                                                                                                                           
									.bytes_allowance                                                                                                                                                                                  
									.saturating_add(bytes_allowance);
```

bkontur: either we do `saturating_add` or change the test(s) https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/24787007601/job/72534708840?pr=448#step:5:68

bkchr: Yeah I have done this on purpose, but if you think adding is more correct I can add it back.

pallets/transaction-storage/src/extension.rs
```diff
@@ -272,3 +274,114 @@ where
...
+/// `BOOST * (bytes_allowance - bytes) / bytes_allowance`. Fresh grant yields the full boost;
+/// at-cap or over-cap yields zero.
+pub const ALLOWANCE_PRIORITY_BOOST: TransactionPriority = 1_000_000_000;
+
+/// Computes `ALLOWANCE_PRIORITY_BOOST * (bytes_allowance - bytes) / bytes_allowance`.
+///
+/// Returns `0` when `bytes_allowance == 0` (missing / expired auth) or when `bytes >=
+/// bytes_allowance` (over cap). Uses `u128` intermediate because
+/// `ALLOWANCE_PRIORITY_BOOST (1e9) * u64_max_plausible_bytes (~1e12)` exceeds `u64::MAX`.
+fn proportional_boost(bytes: u64, bytes_allowance: u64) -> TransactionPriority {
```
karolk91: shouldn't the priority work other way around i.e. 
1. as long as you still have allowance (eg still not used the 100MB limit) -> you get priorty X (so everyone with allowance gets the same priority)
2. when you used all of your allowance - then your priority is starting getting lower depending on how many bytes you stored (getting less than X) ?

current solution could be abused for censorship, eg some person used 50MB and trying to store more but someone other still having 100MB and keep spamming 1 byte txs. With 512 max transactions per block he can keep the everyone from storing anything for 7 days

bkontur: What do you suggest? Take `data.len() and MaxTransactionSize` into account for boost?

karolk91: No, but now I also realized that we don't track currently how much bytes per Account is stored. So I would propose something like below instead of `proportional_boost`:

```
fn boost(bytes: u64, bytes_allowance: u64) -> TransactionPriority {
      if bytes_allowance > 0 && bytes < bytes_allowance {
          ALLOWANCE_PRIORITY_BOOST
      } else {
          0
      }
  }
```

So it means:
1. Everyone who still has allowance has higher priority than someone who used all of his allowance
2. All people with some allowance are equal
3. All people with depleted allowance are equal

bkontur: > we don't track currently how much bytes per Account

Yes, we do with this PR:
```
	/// Bytes consumed so far.
	pub bytes: u64,
```

```
if consume {
        authorization.extent.bytes = authorization.extent.bytes.saturating_add(size.into());
```

But this grows forever, so I added resetting this when `refresh_account_authorization` is triggered here: https://github.com/paritytech/polkadot-bulletin-chain/pull/448/commits/22362fbdbe9e9cfe607b0687adc011b2b1272dda

pallets/transaction-storage/src/lib.rs
bkontur: > Yeah I have done this on purpose, but if you think adding is more correct I can add it back.

Yes, right, I see now, adding is not correct

pallets/transaction-storage/src/extension.rs
bkontur: > No, but now I also realized that we don't track currently how much bytes per Account is stored. So I would propose something like below instead of `proportional_boost`:
> 
> ```
> fn boost(bytes: u64, bytes_allowance: u64) -> TransactionPriority {
>       if bytes_allowance > 0 && bytes < bytes_allowance {
>           ALLOWANCE_PRIORITY_BOOST
>       } else {
>           0
>       }
>   }
> ```
> 
> So it means:
> 
>     1. Everyone who still has allowance has higher priority than someone who used all of his allowance
> 
>     2. All people with some allowance are equal
> 
>     3. All people with depleted allowance are equal


Ok, I put this behind trait, so we can decide later on runtime level, push your suggestion as an `FlatBoost` and keep Basti's as `ProportialBoost`


https://github.com/paritytech/polkadot-bulletin-chain/pull/448/commits/7a3b9dcefd7c37d921573af5c8b5652581315272

bkontur:
@karolk91 @franciscoaguirre  — please take a look at this PR. I’ve tried to minimally align it with [https://github.com/paritytech/individuality/pull/785](https://github.com/paritytech/individuality/pull/785), based on the triangle design from [https://github.com/paritytech/triangle-js-sdks/pull/129](https://github.com/paritytech/triangle-js-sdks/pull/129).

Please review it — I’d like to merge both this PR and its counterpart ([https://github.com/paritytech/individuality/pull/785](https://github.com/paritytech/individuality/pull/785)) and deploy everything to Paseo for testing next week.

This will unblock at least `store`-utilized usage for W3S and beyond.

I also added a boost to the transaction count to prevent spam with 1B transactions, as you both requested 🙂 — I borrowed that part from Cisco’s PR: [https://github.com/paritytech/polkadot-bulletin-chain/pull/469](https://github.com/paritytech/polkadot-bulletin-chain/pull/469).

I’ve prepared changes for the hard cap and renewals as well, but after our discussion I need to think them through more tomorrow: [https://github.com/paritytech/polkadot-bulletin-chain/pull/457](https://github.com/paritytech/polkadot-bulletin-chain/pull/457).

pallets/transaction-storage/src/benchmarking.rs
```diff
@@ -200,13 +200,12 @@ mod benchmarks {
...
 		let who: T::AccountId = whitelisted_caller();
-		let transactions = 10;
 		let bytes: u64 = 1024 * 1024;
 
 		#[extrinsic_call]
-		_(origin as T::RuntimeOrigin, who.clone(), transactions, bytes);
+		_(origin as T::RuntimeOrigin, who.clone(), 0u32, bytes);
 
-		assert_last_event::<T>(Event::AccountAuthorized { who, transactions, bytes }.into());
+		assert_last_event::<T>(Event::AccountAuthorized { who, bytes }.into());
```
franciscoaguirre: Maybe we bring this transactions back?

pallets/transaction-storage/src/extension.rs
```diff
@@ -272,3 +275,235 @@ where
...
+		_inherited_implication: &impl Implication,
+		_source: TransactionSource,
+	) -> ValidateResult<Self::Val, RuntimeCallOf<T>> {
+		let Some(inner_call) = call.is_sub_type() else {
+			return Ok((ValidTransaction::default(), (), origin));
+		};
+		// Only `store` / `store_with_cid_config` get the boost. `renew` also carries
+		// `Origin::Authorized` and does consume allowance, but it operates on
+		// already-stored data and shouldn't compete for the same priority slots as
+		// new submissions.
```
franciscoaguirre, line 405: I actually argue renewals should have the same priority. We can change it in a future PR though after discussion

runtimes/bulletin-westend/tests/tests.rs
```diff
@@ -1018,20 +1148,31 @@ fn signed_store_prefers_preimage_authorization_over_account() {
...
+					transactions_allowance: 0
+				},
 				"Account authorization should remain unchanged when preimage auth is used"
 			);
 		});
 }
 
-/// Renew calls wrapped in utility/sudo require authorization, same as store.
+/// `renew` is allowed as a direct extrinsic (consuming the caller's account allowance)
+/// but rejected when nested inside utility/sudo wrappers.
```
franciscoaguirre, line 1173: We might want to allow this later for batches

franciscoaguirre (approved):

pallets/transaction-storage/src/benchmarking.rs
bkontur: > Maybe we bring this transactions back?

Yes, I had to go babysitting and assistant fixed it mean time: https://github.com/paritytech/polkadot-bulletin-chain/pull/448/commits/1d0f322662662775e4e3f49f95482c03d5741cd2

sorry, I force-push this commit because of adding:
```
Co-authored-by: Francisco Aguirre <franciscoaguirreperez@gmail.com>
```
Because I took those changes from your PR :)

pallets/transaction-storage/src/lib.rs
```diff
@@ -93,13 +93,21 @@ pub const AUTHORIZATION_NOT_FOUND: InvalidTransaction = InvalidTransaction::Cust
...
 
-/// Number of transactions and bytes covered by an authorization.
-#[derive(PartialEq, Eq, Debug, Encode, Decode, scale_info::TypeInfo, MaxEncodedLen)]
+/// Usage state of an authorization. `bytes` / `transactions_used` accumulate upward as
+/// data is stored; `bytes_allowance` / `transactions_allowance` are the caps set at
+/// grant time.
+#[derive(
+	Copy, Clone, PartialEq, Eq, Debug, Default, Encode, Decode, scale_info::TypeInfo, MaxEncodedLen,
+)]
 pub struct AuthorizationExtent {
```
karolk91, line 102: So changes to `AuthorizationExtent` are here only for the reason of `ProportionalBoost` ? I guess we could simplify by a lot if we just decide that we are not going to use `ProportionalBoost`

pallets/transaction-storage/src/lib.rs
```diff
@@ -997,55 +1019,78 @@ pub mod pallet {
...
+		/// counters (`bytes`, `transactions_used`) to `0` so the holder regains a fresh
+		/// allowance for the new period.
 		fn refresh_authorization(scope: AuthorizationScopeFor<T>) -> DispatchResult {
 			let expiration = frame_system::Pallet::<T>::block_number()
 				.saturating_add(T::AuthorizationPeriod::get());
 
 			Authorizations::<T>::mutate(&scope, |maybe_authorization| {
 				if let Some(authorization) = maybe_authorization {
 					authorization.expiration = expiration;
+					// Reset usage so the holder regains a fresh allowance for the new period.
```
karolk91: Not sure - why would refresh give me "additional" authorizations by resetting usage? What would be the flow here ?

pallets/transaction-storage/src/lib.rs
```diff
@@ -1157,57 +1202,41 @@ pub mod pallet {
...
-		/// Check that authorization exists for data of the given size to be stored in a single
-		/// transaction. If `consume` is `true`, the authorization is consumed.
+		/// Check that authorization exists for data of the given size.
+		///
+		/// Rejects only if the authorization entry is missing or expired — never rejects on
+		/// insufficient allowance. If `consume` is `true`, adds `size` to the consumed `bytes`
+		/// counter and increments `transactions_used` by 1 (both saturating); callers can
+		/// overshoot the caps, in which case the [`extension::AllowanceBasedPriority`]
+		/// boost no longer applies.
 		fn check_authorization(
```
karolk91, line 1222: what stops me from spamming unlimited "renew" calls ?

pallets/transaction-storage/src/lib.rs
bkontur: No, your FlatBoost is using those also in [in_budget](https://github.com/paritytech/polkadot-bulletin-chain/pull/448/files#diff-adec689e8bd548e0c86d1dc6ce8d6c82b565ed595e0aa76b29f27703a232594aR296)

pallets/transaction-storage/src/lib.rs
bkontur: > Not sure - why would refresh give me "additional" authorizations by resetting usage? What would be the flow here ?

ah, good catch, thank you, reverted: https://github.com/paritytech/polkadot-bulletin-chain/pull/448/commits/aefb13ca2f31ef042aab1704a50c1ae37e8bed97

this was wrong expectation from previous George PR where we expected that `refresh_authorization` will be used, but it is not needed.

pallets/transaction-storage/src/lib.rs
karolk91: Yes, but to implement `FlatBoost` we don't need to change `AuthorizationExtent` in the first place

pallets/transaction-storage/src/lib.rs
bkontur: > what stops me from spamming unlimited "renew" calls ?

I said above:

> I’ve prepared changes for the hard cap and renewals as well, but after our discussion I need to think them through more tomorrow: https://github.com/paritytech/polkadot-bulletin-chain/pull/457.

bkontur (approved):

pallets/transaction-storage/src/lib.rs
bkontur: > Yes, but to implement `FlatBoost` we don't need to change `AuthorizationExtent` in the first place

But we need to know if the user is "in budget or not", that's why we changed `AuthorizationExtent`, or what do you mean?

PR #457, opened 2026-04-27
Authorizations - hard-cap for renewals

bkontur:
Adds the **hard cap on permanent storage** on top of #448's soft side. (PR #486 has been folded in; this PR lands the simplified accounting design described below.)

Design: [docs/001_authorizations.md](https://github.com/paritytech/polkadot-bulletin-chain/blob/bko-authorizations/docs/001_authorizations.md).

## What this PR adds

- `AuthorizationExtent.bytes_permanent` — per-window renew-quota counter. Increment-only within a window; resets to `0` when the authorization is re-granted on the expired-but-present path.
- Per-account hard cap: `renew` rejected with `Error::PermanentAllowanceExceeded` when `bytes_permanent + size > bytes_allowance`.
- Chain-wide hard cap: `renew` rejected with `Error::ChainPermanentCapReached` when `PermanentStorageUsed + size > T::MaxPermanentStorageSize`.
- `TransactionInfo.kind: TransactionKind { Store, Renew }` discriminator. `on_initialize`'s obsolete-block cleanup decrements `PermanentStorageUsed` by the sum of `kind == Renew` sizes when a block is removed — `Transactions` is itself the source of truth for currently-on-chain renewed bytes.
- `PermanentStorageUsed` chain-wide counter (single source of truth — no separate ledger, no `on_poll` drain, no liveness fallback).
- Events: `PermanentStorageUsedUpdated` (per change), `PermanentStorageNearCap` (rising-edge at 80% of cap).
- `v1→v2` migration: rewrites `AuthorizationExtent` schema; drops zero-quota and expired entries.
- `try_state` invariant: `PermanentStorageUsed == Σ size of renewed Transactions entries`.

## Open

- [ ] Per-content dedup of re-renewals — to be aligned with #313 (`do_renew` is the single point both sides need to converge on).


docs/001_authorizations.md
```diff
@@ -0,0 +1,214 @@
...
+- `authorize_account(who, transactions, bytes)` behaves differently per state of the existing entry:
+    - **Unexpired**: caps are **additive** — `bytes_allowance += bytes`, `transactions_allowance += transactions`. This matches the PoP `claim_long_term_storage` flow, where each successful claim is expected to extend the caps. Consumed counters (`bytes`, `bytes_permanent`, `transactions`) are preserved. Expiry is left untouched; use `refresh_account_authorization` to push it back.
+    - **Expired-but-present**: re-grant the caps (`bytes_allowance = bytes`, `transactions_allowance = transactions`), reset the soft counters to `0` (`bytes = 0`, `transactions = 0`), and **leave `bytes_permanent` as-is**. The lazy ledger drain remains the only path that ever decrements `bytes_permanent`, so in-flight on-chain commitments aren't clobbered by re-authorize. (Pairs with the rule below: `remove_expired_account_authorization` refuses to remove the entry while `bytes_permanent > 0`.)
```
franciscoaguirre: I don't think `bytes_permanent` should be kept between authorization periods after our discussion with @karolk91  😬

docs/001_authorizations.md
```diff
@@ -0,0 +1,214 @@
...
+    - **Missing**: create a fresh entry with all counters at `0`.
+- `authorize_preimage(content_hash, max_size)` follows the same shape but with `transactions_allowance = 1` baked in (a preimage grant is a single-shot store right). Unexpired re-authorize **replaces** the cap (preimage grants are point-in-time, not additive).
+
+### Refresh authorization
+
+`refresh_account_authorization(who)` extends an authorization's lifetime:
+
+- Extends `expiration` by another `AuthorizationPeriod`. Caps and **all** consumed counters (`bytes`, `bytes_permanent`, `transactions`) are left untouched. Refresh does not grant additional capacity.
+- To grant more capacity within an unexpired window, call `authorize_account` (additive on the unexpired path); to fully reset the soft counters, let the authorization expire and re-authorize.
+- `bytes_permanent` survival across refresh is load-bearing: clearing it would let a holder commit unbounded permanent storage by refreshing repeatedly.
```
franciscoaguirre: This is not true since `renew()` takes away from their total allowance per period

docs/001_authorizations.md
bkontur: > I don't think `bytes_permanent` should be kept between authorization periods after our discussion with @karolk91 😬

we don't support periods/slots on Bulletin yet, now we need to keep track of total's user usage to be aligned with hard-cap, there is no other way now

docs/001_authorizations.md
```diff
@@ -0,0 +1,214 @@
...
+Two distinct problems shape the allowance design. They map directly onto the two limits in [Allowance Limits](#allowance-limits).
+
+Block-throughput numbers used below:
+
+| Parameter | Value |
+|---|---|
+| `MaxTransactionSize` | 2 MiB |
+| `MaxBlockTransactions` | 512 |
+| `MAX_BLOCK_LENGTH` × `NORMAL_DISPATCH_RATIO` (90%) | **~9 MiB / block** (binding constraint for storage data) |
+| Block time | 6 s ⇒ 14 400 blocks/day |
```
acatangiu, line 14: Make sure to update this and rest of numbers when we add 3 cores to BC and move to 2s blocks

franciscoaguirre: We won't start with 3 cores as that would need even bigger disks for collators

sdk/metadata.scale
franciscoaguirre, line 1: Remove

pallets/transaction-storage/src/lib.rs
```diff
@@ -911,6 +1003,98 @@ pub mod pallet {
...
+		/// exactly once per crossing.
+		fn update_permanent_storage_used(f: impl FnOnce(u64) -> u64) {
+			let old = PermanentStorageUsed::<T>::get();
+			let new = f(old);
+			PermanentStorageUsed::<T>::put(new);
+			Self::deposit_event(Event::PermanentStorageUsedUpdated { used: new });
+			let cap = T::MaxPermanentStorageSize::get();
+			// Divide-first to avoid u64 overflow on extreme caps (`cap * 80` saturates
+			// above ~230 EiB). Loses ≤`pct` bytes of precision; harmless for the rising-edge.
+			let threshold = (cap / 100).saturating_mul(PERMANENT_STORAGE_NEAR_CAP_PERCENT);
```
franciscoaguirre: Not using `Percent`?

sdk/metadata.scale
bkontur: no we need this, because rust sdk uses that and the rust sdk tests failed because of that

franciscoaguirre (approved):

Issue #431, opened 2026-04-20
[CI] - setup cmd.py bot (bench, fmt, ...)


rosarp:
Available Options:
1. Create CI bot to generate bench-marking
2. <del>Create SKILL to generate bench-marking</del>

bkontur:
> 1. Create CI bot to generate bench-marking

Re-use the same runner as Polkadot SDK uses for bench bot

bkontur:
I think we just need to setup the cmd bot here, we have cmd.py copy (bench, fmt, ...)

danielbui12:
@bkontur @rosarp 
Some clarity questions for bench:
- what happens after benchmarks run — commit back (create another PR merge to current PR) or just report?
- what is the trigger condition (on merge to main, cron, ...)?
Otherwise just follow Polkadot SDK cmd.py bench? https://github.com/paritytech/polkadot-sdk/blob/b1b1139a2876ed992f59722eb043e96a33a176ca/.github/workflows/bench-all-runtimes.yml#L95

For check & fmt, we already have it [here](https://github.com/paritytech/polkadot-bulletin-chain/blob/31b4b7813e6cf0b6e2e20b6ab2236fe8749b0724/.github/workflows/check.yml). what are expectations from this state?

bkontur:
@danielbui12 check PolkadotSDK for bench bot setup.

We usually run the cmd bot (bench, fmt) for the relevant pull requests and yes, the bot will commit/push those changes to the PR directly and yes the only trigger is the user with comment on the PR - e.g. `/cmd fmt` or `/cmd bench --runtime....

danielbui12:
Ref https://github.com/paritytech/polkadot-sdk/blob/b1b1139a2876ed992f59722eb043e96a33a176ca/.github/workflows/cmd.yml

PR #463, opened 2026-04-27
claude skill to perform benchmarking for a runtime

rosarp:
Does similar to #431 but as a claude skill.


.claude/skills/benchmarking/bench-runner.sh
```diff
@@ -0,0 +1,133 @@
+#!/usr/bin/env bash
+# Reusable runtime-benchmark runner.
```
bkontur: @rosarp I would not go with dedicated script for Claude, why not just tell Clade to use `python3 ./scripts/cmd/cmd.py bench...`?

rosarp: @bkontur there were some caveats, but can be done other way.
will fix this and remove bench-runner.sh

PR #471, opened 2026-04-30
Prepare for TransactionStorageAPI v2

skunert:
Since https://github.com/paritytech/polkadot-sdk/pull/11939 is merged, this PR prepares the bulletin side.

Main changes:
- Transactions storage item now contains `extrinsic_index`
- Transaction storage item `size` field is now public so that we can construct the runtime API return struct which lives in polkadot-sdk
- Most of the code here is just tests & migration related code. This is only needed for paseo & westend to need to reset. For polkadot we should directly start with the fresh state.
- `transactions_at` can decode v2 and v3 format, so that the runtime API can be used while a MBM is running. This might not be strictly necessary, but allows us to test syncing also on paseo & westend.


pallets/transaction-storage/src/migrations.rs
```diff
@@ -362,3 +362,160 @@ pub mod v2 {
...
+		pub hashing: HashingAlgorithm,
+		pub cid_codec: CidCodec,
+		pub size: u32,
+		pub block_chunks: ChunkIndex,
+	}
+
+	/// Stepped migration from storage version 2 to 3.
+	pub struct MigrateV2ToV3<T: Config>(PhantomData<T>);
+
+	impl<T: Config> SteppedMigration for MigrateV2ToV3<T> {
```
skunert, line 405: Hmm now I think this migration is not even needed. If we read version agnostic, the state will clean itself up after 2 weeks. Will remove it later.

pallets/transaction-storage/src/lib.rs
```diff
@@ -1216,6 +1223,42 @@ pub mod pallet {
 			transactions.into_iter().nth(index as usize)
 		}
 
+		/// All transactions stored at the given block, in the current `TransactionInfo` layout.
+		///
+		/// Shape-tolerant against entries that are still in the pre-v3 layout.
+		pub fn transactions_at(
```
bkontur, line 1233: @skunert leaving TODOs for adding this: https://github.com/paritytech/polkadot-bulletin-chain/pull/476/files#diff-deb38b991055ff37e7403027e6b043e18d07d49f3d0345f30780abe06018a696R882

pallets/transaction-storage/src/migrations.rs
skunert: I left it in in the end because Paseo should be as close to Polkadot as possible. And without the migration, performance would be different when accessing old items.

skunert:
@bkontur @karolk91 This is now ready for review. Note that the migration is only necessary for Paseo/Westend. I ran the migration on current paseo state with 5761 entries via try-runtime and it took 49 blocks.

pallets/transaction-storage/src/lib.rs
skunert: Implemented them

runtimes/bulletin-westend/src/lib.rs
```diff
@@ -390,6 +391,25 @@ impl parachain_info::Config for Runtime {}
...
+		Perbill::from_percent(50) * RuntimeBlockWeights::get().max_block;
+}
+
+impl pallet_migrations::Config for Runtime {
+	type RuntimeEvent = RuntimeEvent;
+	#[cfg(not(feature = "runtime-benchmarks"))]
+	type Migrations = migrations::MbmMigrations;
+	#[cfg(feature = "runtime-benchmarks")]
+	type Migrations = pallet_migrations::mock_helpers::MockedMigrations;
+	type CursorMaxLen = ConstU32<65_536>;
```
bkontur, line 405: @skunert not sure how much this number matter, but we won't have situation that we need to migrated 2*weeks of blocks full with 512 transactions

bkontur (approved):
@skunert looks good to me, what about that failed check migration: https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/25363525457/job/74368706142?pr=471#step:8:122?

bkontur:
/cmd bench --runtime bulletin-paseo --pallet pallet_bulletin_transaction_storage

bkontur:
/cmd bench --runtime bulletin-westend --pallet pallet_bulletin_transaction_storage

skunert:
> @skunert looks good to me, what about that failed check migration: https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/25363525457/job/74368706142?pr=471#step:8:122?


looking into it

skunert:
Fixed, problem was that inherent was pruning Transactions during MBM and this broke the try-runtime test. Because we were assuming the count does not change during the migration.

bkontur:
> Fixed, problem was that inherent was pruning Transactions during MBM and this broke the try-runtime test. Because we were assuming the count does not change during the migration.

nice, thank you.

@karolk91 do you want to do also final check before we merge?

pallets/transaction-storage/src/lib.rs
```diff
@@ -177,7 +177,12 @@ pub struct TransactionInfo {
 	pub cid_codec: CidCodec,
 
 	/// Size of indexed data in bytes.
-	size: u32,
+	pub size: u32,
+	/// Extrinsic index within the block that originally indexed this data
+	/// (via `sp_io::transaction_index::index` / `renew`). For renewed entries
+	/// this is the renewer's extrinsic index, not the original. Entries created
+	/// before storage version 2 have `u32::MAX` as a sentinel.
```
karolk91: Just 2 nits:
```suggestion
	/// before storage version 3 have `u32::MAX` as a sentinel.
```

And seems there is a case that `renew` would fail during migration because it uses `Transactions::<T>::get` vs `Self::transactions_at`

Both above are really not important so feel free to ignore!

karolk91 (approved):

pallets/transaction-storage/src/lib.rs
bkontur:
```suggestion
	/// this is the renewer's extrinsic index, not the original.
```

bkontur: I removed that part `/// before storage version 3 have `u32::MAX` as a sentinel.`, those will gets prunned meantime,
and also this migration is relevant just for Paseo/Westend, where we probably don't even have renewals triggered

skunert: During MBM no normal transactions can go live, only inherents. So during this time we will have no renew and no store.

PR #485, opened 2026-05-05
fix: add pre-install frame-omni-bencher for cmd bot

danielbui12:
Reminder: restore branch after merge


scripts/cmd/cmd.py
```diff
@@ -91,6 +145,8 @@ def setup_logging():
 print_and_log(f'args: {args}')
 
 if args.command == 'bench':
+    install_frame_omni_bencher()
```
bkontur: @danielbui12 why? Is this the same for PolkadotSDK?

danielbui12: @bkontur in polkadot-sdk, frame-omni-bencher is built from source

bkontur: > @bkontur in polkadot-sdk, frame-omni-bencher is built from source

but we don't need and want to build it from source here, no reason for that. PolkadotSDK builds from source because it also tests bencher itself. Here we just download and cache one.

scripts/cmd/cmd.py
```diff
@@ -22,7 +24,59 @@ def setup_logging():
...
+
+    if not version:
+        print_and_log('❌ POLKADOT_SDK_VERSION not set in environment')
+        sys.exit(1)
+
+    install_path = '/usr/local/bin/frame-omni-bencher'
+    url = f'https://github.com/paritytech/polkadot-sdk/releases/download/{version}/frame-omni-bencher'
+    print_and_log(f'Downloading frame-omni-bencher {version} from {url}...')
+
+    if os.system(f'curl -fL -o frame-omni-bencher {url}') != 0:
```
danielbui12: @bkontur I only put the download script, not clone polkadot sdk then build from source here

scripts/cmd/cmd.py
danielbui12: @bkontur agree, please check this https://github.com/paritytech/polkadot-bulletin-chain/pull/485#discussion_r3194160937

scripts/cmd/cmd.py
bkontur: > @bkontur I only put the download script, not clone polkadot sdk then build from source here

but we download and cache the bencher in CI, so this will be double-download :)

If we really need `install_frame_omni_bencher` I would make it conditional, and install_frame_omni_bencher should take the argument POLKADOT_SDK_VERSION instead of env var `version = os.environ.get('POLKADOT_SDK_VERSION')`

and this way, it would install only if you pass it: `/cmd bench --bencher--sdk-version stable2603`.

What are you trying to solve with this install? `cmd.yml / cmd-run.yml`  does not preinstall bencher? If not, ok, you can read the version from `env` and pass `/cmd bench --bencher-sdk-version ${POLKADOT_SDK_VERSION}` or download and cache bencher in CI as we do in check.yml:
```
      - name: Cache frame-omni-bencher
      - name: Download frame-omni-bencher
```

danielbui12: @bkontur 
I move the `frame-omni-bencher` downloading to CI.

First time download `frame-omni-bencher`: https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/25536193285/job/74952447309
Next time, CI cached it: https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/25538161174/job/74958247766

.github/workflows/cmd-run.yml
```diff
@@ -172,6 +172,29 @@ jobs:
...
+          key: frame-omni-bencher-${{ env.POLKADOT_SDK_VERSION }}
+
+      - name: Download frame-omni-bencher
+        if: startsWith(github.event.inputs.cmd, 'bench') && steps.cache-bencher.outputs.cache-hit != 'true'
+        run: |
+          url="https://github.com/paritytech/polkadot-sdk/releases/download/${POLKADOT_SDK_VERSION}/frame-omni-bencher"
+          curl -fL -o /tmp/frame-omni-bencher "$url"
+          if [ -n "${FRAME_OMNI_BENCHER_SHA256}" ]; then
+            echo "${FRAME_OMNI_BENCHER_SHA256}  /tmp/frame-omni-bencher" | sha256sum -c
+          fi
```
bkontur:
```suggestion
          echo "${FRAME_OMNI_BENCHER_SHA256} /tmp/frame-omni-bencher" | sha256sum -c -
```

bkontur (approved):

PR #486, opened 2026-05-05
Simplify renewal hard cap

franciscoaguirre:
Instead of giving back `bytes_permanent` to accounts, we can rely on their ephemeral nature to cap renewals: an account needs to constantly renew for a blob to be kept. Every time they renew they have X amount of `bytes_allowance` they can use. If they want to renew something forever (akin to permanent storage) they can do so using their same X allowance.

To maintain a global limit on data being renewed I added a new field to `TransactionInfo` that keeps track of whether a blob came from `Store` or `Renew`, if it came from renew we increase the global accounting and decrease it when that blob is pruned in `on_initialize` one `RetentionPeriod` later (if not renewed again).


pallets/transaction-storage/src/lib.rs
```diff
@@ -392,45 +409,34 @@ pub mod pallet {
...
+			// renewed bytes that just aged out. The proof for `obsolete` will be checked
+			// later in this block, so we drop `obsolete` - 1.
 			weight.saturating_accrue(db_weight.reads(1));
 			let period = Self::retention_period();
 			let obsolete = n.saturating_sub(period.saturating_add(One::one()));
 			if obsolete > Zero::zero() {
-				weight.saturating_accrue(db_weight.writes(1));
-				<Transactions<T>>::remove(obsolete);
+				weight.saturating_accrue(db_weight.reads_writes(1, 1));
+				if let Some(removed) = <Transactions<T>>::take(obsolete) {
```
bkontur, line 424: @karolk91 @franciscoaguirre relates to https://github.com/paritytech/polkadot-bulletin-chain/pull/313/files#r3186973298

pallets/transaction-storage/src/migrations.rs
```diff
@@ -344,17 +372,82 @@ pub mod v2 {
...
 			);
-			// One read + one write per visited entry (translate rewrites or deletes).
-			let touched = migrated.saturating_add(dropped);
-			T::DbWeight::get().reads_writes(touched, touched)
+
+			// Tail-extend every TransactionInfo with `kind = Store`. Pre-migration
+			// renewed bytes age out without bumping/decrementing the chain counter,
+			// which starts at 0 and converges within one RetentionPeriod.
+			let mut tx_blocks_migrated: u64 = 0;
+			Transactions::<T>::translate::<BoundedVec<V1TransactionInfo, T::MaxBlockTransactions>, _>(
```
bkontur: @franciscoaguirre this won't work for Paseo, we need to do it as a SteppedMigration - but we will join it here: https://github.com/paritytech/polkadot-bulletin-chain/pull/471/files#diff-5e571bb2ee729e52d0593bb641ddd04b46bc9fa2bc5f26f5bbf7ff59b2bacc72R405  (no need to fix now)

pallets/transaction-storage/src/migrations.rs
```diff
@@ -344,17 +358,59 @@ pub mod v2 {
...
 			);
-			// One read + one write per visited entry (translate rewrites or deletes).
-			let touched = migrated.saturating_add(dropped);
-			T::DbWeight::get().reads_writes(touched, touched)
+
+			// `Transactions` is intentionally not rewritten here — `v2→v3` decodes v1
+			// entries via its `V2TransactionInfo` path (same SCALE shape) and converts
+			// them in bounded per-block steps. See the module-level doc comment.
+
+			// `BlockTransactions` is transient (cleared in `on_finalize`) so it's
```
bkontur, line 370: not sure if this could even happen, but let's keep

bkontur (approved):
@franciscoaguirre good job :)

PR #491, opened 2026-05-05
Add runtime API for authorizations

franciscoaguirre:
Addresses https://github.com/paritytech/polkadot-bulletin-chain/issues/495

We have recently changed the `AuthorizationExtent` fields with data irrelevant for most client applications. I'm adding a runtime API for returning a simpler and more stable version of authorizations for client applications to use.

## TODO
- [x] Add expiration


pallets/transaction-storage/runtime-api/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+use scale_info::TypeInfo;
+
+/// Remaining storage capacity for an account.
+///
+/// All fields are zero when the account has no authorization, has fully
+/// consumed it, or its authorization has expired.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Encode, Decode, TypeInfo)]
+pub struct AccountStorageAuthorization {
+	/// Remaining number of store/renew transactions the account may submit.
+	pub transactions: u32,
```
bkontur: I would remove transactions, not used and we dont block on them

```suggestion
```

franciscoaguirre: We might renew renewals on it but you're right if you use things correctly you would never have to worry about this

pallets/transaction-storage/runtime-api/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+	/// Remaining bytes the account may store.
+	pub bytes: u64,
```
bkontur: @franciscoaguirre what about adding those bytes / bytes_allowance / bytes_permanent, so the client can decide, hmm but also just `bytes` could be easy - and we return/implement  `bytes_allowance - bytes_permanent` ?

pallets/transaction-storage/runtime-api/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "pallet-bulletin-transaction-storage-runtime-api"
```
bkontur, line 2: @franciscoaguirre new module, we need to also add it to the publish-crates.yml for crates.io releases

runtimes/bulletin-paseo/src/lib.rs
```diff
@@ -986,6 +986,20 @@ impl_runtime_apis! {
...
+	impl pallet_bulletin_transaction_storage_runtime_api::TransactionStorageAuthorizationApi<Block, AccountId, BlockNumber> for Runtime {
+		fn account_authorization(
+			account: AccountId,
+		) -> Option<pallet_bulletin_transaction_storage_runtime_api::AccountAuthorization<BlockNumber>> {
+			let expires_at = pallet_bulletin_transaction_storage::Pallet::<Runtime>::account_authorization_expires_at(account.clone())?;
+			let extent = pallet_bulletin_transaction_storage::Pallet::<Runtime>::account_authorization_extent(account);
+			Some(pallet_bulletin_transaction_storage_runtime_api::AccountAuthorization {
+				expires_at,
+				priority_bytes: extent.bytes_allowance.saturating_sub(extent.bytes),
+				renew_bytes: extent.bytes_allowance,
```
bkontur: doc says "/// Bytes available for `renew` calls (`bytes_allowance - bytes_permanent`)."

We should hide these calculations in pallet call, so we don't need to do this for every runtime

pallets/transaction-storage/runtime-api/src/lib.rs
```diff
@@ -0,0 +1,39 @@
...
+/// - **Is the authorization still valid?** — `expires_at` is the block at which it lapses; the API
+///   returns `None` for accounts with no active authorization.
+/// - **Will my next `store` call be high-priority?** — yes, iff `size <= priority_bytes` (and a
+///   transaction slot is still available). Stores that exceed `priority_bytes` still validate, just
+///   at base priority.
+/// - **Can I renew?** — yes, iff `size <= renew_bytes`.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Encode, Decode, TypeInfo)]
+pub struct AccountAuthorization<BlockNumber> {
+	/// Block at which this account's authorization expires.
+	pub expires_at: BlockNumber,
```
bkontur, line 19: not sure if client will be able to interpret this - relay block, Bulletin para block?

pallets/transaction-storage/runtime-api/src/lib.rs
```diff
@@ -0,0 +1,39 @@
...
+	/// Bytes that fit inside the high-priority window for `store` calls
+	/// (`bytes_allowance - bytes`).
+	pub priority_bytes: u64,
+	/// Bytes available for `renew` calls (`bytes_allowance - bytes_permanent`).
+	pub renew_bytes: u64,
```
bkontur: Either we will do `substract` and return "free" bytes (soft, hard)

```suggestion
    /// Quota
    pub bytes_allowance: u64,
    /// Bytes that fit inside the high-priority window for `store` calls
	/// (`bytes_allowance - bytes`).
	pub bytes_free: u64,
	/// Bytes available for `renew` calls (`bytes_allowance - bytes_permanent`).
	pub bytes_permanent_free: u64,
```

or we just return values, we don't need to do substract and client can calculate "free".

```suggestion
    /// Quota
    pub bytes_allowance: u64,
    /// Bytes that fit inside the high-priority window for `store` calls
	/// (`bytes`).
	pub bytes_used: u64,
	/// Bytes already used for `renew` calls (`bytes_permanent`).
	pub bytes_permanent_used: u64,
```

runtimes/bulletin-paseo/src/lib.rs
```diff
@@ -986,6 +986,20 @@ impl_runtime_apis! {
 		}
 	}
 
+	impl pallet_bulletin_transaction_storage_runtime_api::TransactionStorageAuthorizationApi<Block, AccountId, BlockNumber> for Runtime {
+		fn account_authorization(
+			account: AccountId,
+		) -> Option<pallet_bulletin_transaction_storage_runtime_api::AccountAuthorization<BlockNumber>> {
+			let expires_at = pallet_bulletin_transaction_storage::Pallet::<Runtime>::account_authorization_expires_at(account.clone())?;
+			let extent = pallet_bulletin_transaction_storage::Pallet::<Runtime>::account_authorization_extent(account);
```
bkontur: I think both are doing `Authorizations::<T>::get(AuthorizationScope::Account(who))?`, I would create dedicated function, returning the relevant data, so we don't need to duplicate this code in every runtime 

```suggestion
			let expires_at = pallet_bulletin_transaction_storage::Pallet::<Runtime>::account_authorization_expires_at(account.clone())?;
			let extent = pallet_bulletin_transaction_storage::Pallet::<Runtime>::account_authorization_extent(account);
```

pallets/transaction-storage/runtime-api/src/lib.rs
```diff
@@ -0,0 +1,39 @@
...
+}
+
+sp_api::decl_runtime_apis! {
+	/// Runtime API for querying transaction-storage authorizations.
+	pub trait TransactionStorageAuthorizationApi<AccountId, BlockNumber>
```
bkontur: @franciscoaguirre we already have one `TransactionStorageApi` (for retention_period and proofs, sync stuff), so I would rename this `TransactionStorageAuthorizationApi` -> `BulletinTransactionStorageApi` - it won't be just about authorizations, I have a feeling that we will need to more APIs here, not just related to the authorizations

```suggestion
	/// Bulletin Runtime API.
	pub trait BulletinTransactionStorageApi<AccountId, BlockNumber>
```

pallets/transaction-storage/runtime-api/src/lib.rs
franciscoaguirre: Do you want to use time? Relay chain block would be better for me, I'm vouching for that

bkontur: > Do you want to use time? Relay chain block would be better for me, I'm vouching for that

Well, I don't know now, we don't use relay chain blocks, and everything is in Bulletin parachain blocks. I am not sure about the time. For now, with actual impl, I would just consider this Bulletin para block number and client could interpret this to the real clock time?

franciscoaguirre: With [authorization slots](https://github.com/paritytech/polkadot-bulletin-chain/pull/509), this will be relay blocks

pallets/transaction-storage/src/lib.rs
```diff
@@ -1757,6 +1758,83 @@ pub mod pallet {
 			Self::authorization_extent(AuthorizationScope::Account(who))
 		}
 
+		/// Returns the expiration block of the account's authorization, or `None` if
+		/// the account has no authorization or its only authorization has expired.
+		pub fn account_authorization_expires_at(who: T::AccountId) -> Option<BlockNumberFor<T>> {
+			let authorization = Authorizations::<T>::get(AuthorizationScope::Account(who))?;
+			(!authorization.expired(Self::now())).then_some(authorization.expiration)
+		}
+
```
bkontur: looks like unused

```suggestion
```

pallets/transaction-storage/src/lib.rs
```diff
@@ -1757,6 +1758,83 @@ pub mod pallet {
...
+		///
+		/// - `content_hash` refers to currently-stored data
+		/// - the stored data's size is within `[1, MaxTransactionSize]`
+		/// - `who` has an unexpired authorization entry
+		/// - per-account hard cap: `bytes_permanent + size <= bytes_allowance`
+		/// - chain-wide hard cap: `PermanentStorageUsed + size <= MaxPermanentStorageSize`
+		pub fn can_renew(who: &T::AccountId, content_hash: ContentHash) -> bool {
+			let Some((block, index)) = TransactionByContentHash::<T>::get(content_hash) else {
+				return false;
+			};
```
bkontur: I am just thinking, that for `fn renew` and `fn force_renew` we use `entry: &TransactionRef`, so maybe it would make sense to use this also here, to cover more cases. Maybe we should also change `fn enable_auto_renewal`:

```suggestion
		pub fn can_renew(who: &T::AccountId, entry: &TransactionRef<BlockNumberFor<T>>) -> bool {
			let Ok(info) = Pallet<T>::resolve_transaction_ref(entry) else { return false };
```

pallets/transaction-storage/src/lib.rs
```diff
@@ -1757,6 +1758,83 @@ pub mod pallet {
...
+		///
+		/// Mirrors the preconditions enforced by [`Self::renew`] +
+		/// [`Self::check_authorization`] (`is_renew = true`):
+		///
+		/// - `content_hash` refers to currently-stored data
+		/// - the stored data's size is within `[1, MaxTransactionSize]`
+		/// - `who` has an unexpired authorization entry
+		/// - per-account hard cap: `bytes_permanent + size <= bytes_allowance`
+		/// - chain-wide hard cap: `PermanentStorageUsed + size <= MaxPermanentStorageSize`
+		pub fn can_renew(who: &T::AccountId, content_hash: ContentHash) -> bool {
```
karolk91: I think `can_renew` should also take into account the case of `AUTO_RENEWAL_ALREADY_ENABLED` - maybe we could extract some common "verify" method between this API and `check_signed` (related to renews) so these are less likely to diverge

karolk91:
If we are going the direction of `can_renew` and `can_store`, I'm afraid that the most common case will be the need to check against multiple items instead of every single one in separation - even if we can check every item separately, especially in case of can_renew, that will not answer the question if its possible to renew batch of items

bkontur:
> If we are going the direction of `can_renew` and `can_store`, I'm afraid that the most common case will be the need to check against multiple items instead of every single one in separation - even if we can check every item separately, especially in case of can_renew, that will not answer the question if its possible to renew batch of items

`can_store` takes bytes, so the client just `sum` all the chunks -> so should be ok, just one call.

hmm, but `can_renew`, good question, I mean I am ok to merge as it is now, `can_renew(one_tx)`.

if we go with `BulletinTransactionStorageApi::can _renew_batch(batch: Vec<TransactionRef>` we need to resolve the nice result type `List<Result<(), Error>` or something like this + maybe limit the input Vec cize? it can be a follow-up.

@karolk91 what do you suggest?

bkontur (approved):

karolk91:
> > If we are going the direction of `can_renew` and `can_store`, I'm afraid that the most common case will be the need to check against multiple items instead of every single one in separation - even if we can check every item separately, especially in case of can_renew, that will not answer the question if its possible to renew batch of items
> 
> `can_store` takes bytes, so the client just `sum` all the chunks -> so should be ok, just one call.
> 
> hmm, but `can_renew`, good question, I mean I am ok to merge as it is now, `can_renew(one_tx)`.
> 
> if we go with `BulletinTransactionStorageApi::can _renew_batch(batch: Vec<TransactionRef>` we need to resolve the nice result type `List<Result<(), Error>` or something like this + maybe limit the input Vec cize? it can be a follow-up.
> 
> @karolk91 what do you suggest?

Yeah I was thinking about something similar, to allow `Vec` in the `can_renew`. But would need to double check what are the possible limits here - how big this `Vec` could be. We could do some follow-up but current real life scenarios always involve multiple CIDs - no one will use `can_renew` or `can_store` and just use `account_authorization` instead


In addition to above, current APIs also doesn't support preimage authorizations

bkontur:
> Yeah I was thinking about something similar, to allow `Vec` in the `can_renew`. But would need to double check what are the possible limits here - how big this `Vec` could be. We could do some follow-up but current real life scenarios always involve multiple CIDs - no one will use `can_renew` or `can_store` and just use `account_authorization` instead

Yes, agree, let's iterate


 
> In addition to above, current APIs also doesn't support preimage authorizations

Well, yes, good point, but who can `auhorize_preimage`? Only OpenGov now, we don't have use-cases for that, again, we can iterate with follow-ups also.

franciscoaguirre:
@bkontur I was thinking about ideas for the runtime API. I will make a new PR then :)

PR #497, opened 2026-05-06
Regression tests for PC -> BC authorize_account scenarios

x3c41a:
People Chain sends an XCM `Transact(authorize_account)` to Bulletin Chain to grant a user a storage allowance. Bulletin Chain had no tests for this path. So a change to the pallet, the XCM config, or the call filters **could break it** without anyone noticing.

Adds 14 tests that send those messages from a simulated People Chain and check what happens on Bulletin Chain. Each scenario lives in its own file under `runtimes/bulletin-paseo/tests/pc_xcm_integration/`:

**`authorize_semantics.rs`**
- one user, single grant from People Chain, fresh state
- two users' allowances on Bulletin Chain are independent
- two grants from People Chain in one period add up on Bulletin Chain; expiry stays where it was
- after expiry, the next grant resets the counters on Bulletin Chain and replaces the limits

**`origin_rejections.rs`**
- a relay chain origin, an `OriginKind::SovereignAccount` from People Chain, and a random local origin all fail to authorize on Bulletin Chain
- `bytes: 0` is rejected on Bulletin Chain

**`safe_call_filter.rs`**
- Bulletin Chain's `SafeCallFilter` blocks `store` over XCM, including inside `Utility::batch`; a batch of only `authorize_account` calls still goes through

**`refresh.rs`**
- `refresh_account_authorization` from People Chain only extends expiry on Bulletin Chain; allowances and consumed counters do not change
- `refresh_account_authorization` fails on Bulletin Chain without a prior authorize

**`end_to_end.rs`**
- People Chain calls `authorize_account`, then the user submits a signed `store` on Bulletin Chain, then a signed `renew`

Closes #490

PR #501, opened 2026-05-07
test-utils: extract XCM authorize_account conformance suite

x3c41a:
Follow-up to #497.

Moves 13 of the 14 People Chain → Bulletin Chain XCM integration tests into a new `bulletin-runtimes-test-utils` crate as generic functions over the runtime and its `XcmConfig`. The authorizer's `Location` is a parameter, so the same suite covers People Chain today and any future authorizer (Fellows, others) without duplication.

The Paseo runtime calls them as thin wrappers under `runtimes/bulletin-paseo/tests/pc_xcm_integration/`, one file per scenario.

The end-to-end test (XCM authorize → signed `store` → signed `renew`) stays runtime-local because it constructs signed extrinsics with Paseo's tx-extension stack.

Per review:

- Lives in its own crate (`runtimes/test-utils`) rather than as a feature on `bulletin-pallets-common`, to keep production and test surfaces separate and avoid feature-propagation churn.
- Authorizer is a `Location` parameter; helper and test names are authorizer-agnostic.

Stacked on `ndk/test-pc-authorize-account-490`. Should merge after #497.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2016,6 +2016,7 @@ name = "bulletin-paseo-runtime"
 version = "0.1.0"
 dependencies = [
  "bulletin-pallets-common",
+ "bulletin-runtimes-test-utils",
  "bulletin-transaction-storage-primitives",
  "cumulus-pallet-aura-ext",
  "cumulus-pallet-parachain-system",
@@ -2083,6 +2084,22 @@ dependencies = [
  "substrate-wasm-builder",
  "tracing",
  "xcm-runtime-apis",
+]
+
+[[package]]
+name = "bulletin-runtimes-test-utils"
+version = "0.1.0"
+dependencies = [
+ "frame-system",
+ "pallet-bulletin-transaction-storage",
+ "pallet-utility",
+ "parachains-common",
+ "parity-scale-codec",
+ "polkadot-sdk-frame",
+ "sp-io",
+ "sp-keyring",
+ "staging-xcm",
+ "staging-xcm-executor",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ westend-runtime-constants = { version = "27.0.0", package = "westend-runtime-con
 # Local workspace members
 bulletin-pallets-common = { version = "0.1.0", path = "pallets/common", default-features = false }
 bulletin-paseo-runtime = { path = "runtimes/bulletin-paseo" }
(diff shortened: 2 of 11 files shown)


pallets/common/src/lib.rs
```diff
@@ -17,6 +17,9 @@
 
 extern crate alloc;
 
+#[cfg(feature = "test-helpers")]
```
franciscoaguirre: I don't like the idea of using this crate for both helpers for production and also test helpers. Having this feature here might cause problems with feature propagation, which we have had a few in the past. I would move these test helpers to a separate crate that runtimes can depend on on `dev-dependencies` and that's it. `bulletin-runtimes-test-utils` or something like that

pallets/common/Cargo.toml
```diff
@@ -39,6 +51,23 @@ std = [
...
+	"dep:xcm-executor",
+	"frame-system/std",
+	"pallet-bulletin-transaction-storage/std",
+	"parachains-common/std",
+	"sp-io/std",
+	"sp-keyring/std",
+	"std",
+	"xcm-executor/std",
+	"xcm/std",
+]
```
franciscoaguirre: The complexity starts to creep here

pallets/common/src/test_helpers.rs
```diff
@@ -0,0 +1,607 @@
...
+// ============================================================================
+// Conformance tests
+//
+// Naming: each public function corresponds to one test scenario. Runtimes
+// wrap them in `#[test]` items and supply `pc_location`, an externalities
+// builder, and an `advance_block` callback that runs the runtime's
+// initialize/finalize hooks.
+// ============================================================================
+
+/// Scenario 1 — happy path. A sibling parachain authorizes a single account
```
franciscoaguirre: I wouldn't put these scenario numbers

pallets/common/src/test_helpers.rs
```diff
@@ -0,0 +1,607 @@
...
+/// from a fresh state; the stored extent matches the granted caps and the
+/// entry is active.
+pub fn pc_authorize_happy_path<R, Cfg>(
```
franciscoaguirre: Also, all of these tests are also valid for any authorizer, not just pc. Maybe we could have the origin be a parameter?

pallets/common/src/lib.rs
x3c41a: Maybe we should not focus on this follow up now but first resolve the issues and merge the base PR; will see if this one is needed...

PR #502, opened 2026-05-07
Nits for ar and authorizations

bkontur:


franciscoaguirre (approved):

x3c41a (approved):

pallets/transaction-storage/src/lib.rs
```diff
@@ -132,6 +132,13 @@ pub struct AuthorizationExtent {
 	pub bytes_allowance: u64,
 }
 
+impl AuthorizationExtent {
+	/// Per-account renew quota check: `bytes_permanent + size <= bytes_allowance`.
+	pub fn renew_fits(&self, size: u64) -> bool {
+		self.bytes_permanent.saturating_add(size) <= self.bytes_allowance
```
franciscoaguirre, line 138: Wait shouldn't this be `self.bytes + self.bytes_permanent + size <= self.bytes_allowance`? Else we are giving people double the allowance, one time for storage and another time for renewals.

bkontur: `self.bytes_permanent` a.k.a renewals are counted for hard-cap, `self.bytes` is for `store` limited by block space - no need to `self.bytes + self.bytes_permanent` sum them up here. We use this check on several places `renew_fits` if just extraction/dedup

bkontur:
/cmd bench --runtime bulletin-paseo bulletin-westend --pallet pallet_bulletin_transaction_storage

pallets/transaction-storage/src/lib.rs
```diff
@@ -132,6 +132,13 @@ pub struct AuthorizationExtent {
 	pub bytes_allowance: u64,
 }
 
+impl AuthorizationExtent {
+	/// Per-account renew quota check: `bytes_permanent + size <= bytes_allowance`.
+	pub fn renew_fits(&self, size: u64) -> bool {
```
karolk91:
```suggestion
	pub fn has_permanent_capacity(&self, size: u64) -> bool {
```
`renew` doesn't seem like a good fit here - name should be more generic

bkontur: ok, I like `has_permanent_capacity`

karolk91 (approved):

PR #504, opened 2026-05-07
Make transaction_storage::enable_auto_renew feeless and authorization-paid

rosarp:
Makes the renew family feeless and authorization-paid end-to-end, and reshapes the surface around a unified `TransactionRef` enum. Mirrors what #474 did for the legacy `renew`; extends the same treatment to `enable_auto_renew`, `disable_auto_renew`, and a new `renew` one-shot scheduler.

#### What this PR changes

- **Renames**: legacy synchronous `renew(block, index)` → `force_renew(entry: TransactionRef)`. The new parameter accepts either `Position { block, index }` or `ContentHash(_)`.
- **New extrinsic**: `renew(entry: TransactionRef)` — feeless **one-shot scheduler with pre-paid renewal**. Charges `bytes_permanent`, `PermanentStorageUsed`, and one tx slot at registration (same as `force_renew`); writes `AutoRenewals[hash] = { recurring: false }`. The renewal fires once at the next `RetentionPeriod` boundary, **without re-charging auth** at cycle time — the slot is already paid for. Pre-payment caps spam: a caller cannot over-schedule past `bytes_allowance`.
- **Removed**: `renew_content_hash`. Folded into `force_renew(TransactionRef::ContentHash(_))` for synchronous renewals; the by-content-hash scheduler variant lives in the unified `renew(entry)`.
- **`enable_auto_renew` is now feeless**. Pool-level snapshot check + one tx slot at registration; `bytes_permanent`, `PermanentStorageUsed`, and another tx slot are charged per cycle by `do_process_auto_renewals`. Origin: `ensure_authorized` (rewritten by the extension to `Origin::Authorized` for account-scoped signed calls; Root/Unsigned rejected).
- **`disable_auto_renew` is now feeless**. `check_signed` gates pool admission on ownership (`AutoRenewals[hash].account == who`), so spam is bounded structurally. Root override available for governance cleanup.
- **`force_renew` accepts both `Position { block, index }` and `ContentHash(_)`** via `TransactionRef`. Same accounting either way; the content-hash variant pays one extra `TransactionByContentHash` read.
- **`AutoRenewalData → RenewalData`** with a new `recurring: bool` field. `STORAGE_VERSION` bumps to 4; MBM `v4::MigrateV3ToV4` re-encodes existing entries with `recurring: true` (every pre-existing entry was written by the old `enable_auto_renew`).
- **Event renamed**: `AutoRenewalEnabled` → `RenewalEnabled { content_hash, who, recurring }`. `AutoRenewalDisabled.who` is the registration owner (not the caller in the Root override path).
- **Latest-entry guard in `on_initialize`** skips obsolete entries that are no longer the latest reference — handles the case where a user-initiated `force_renew` moves `TransactionByContentHash` forward between registration and cycle.
- **Pool tag prefixes split** so `store` and renew operations don't dedup against each other in the pool: `"TransactionStorageStore"` vs `"TransactionStorageRenew"`.
- **Centralized accounting**: hard-cap checks live in `check_authorization` (extension's `check_signed` for the manual flow; auto-renewal drain for the cycle flow). Renewal mechanics live in `do_renew_in_memory`, used by both `do_renew` and `do_process_auto_renewals`.

#### What this PR does not change

- The chain-wide cap (`MaxPermanentStorageSize`) and the `RetentionPeriod` lifecycle model.
- `do_process_auto_renewals` is still mandatory and shares `BlockTransactions` slots with user txs (slot reservation is open follow-up — see `docs/001_authorizations.md` TODO).
- `enable_auto_renew`'s snapshot check is still non-consuming; over-scheduling recurring auto-renewals fails at cycle time with `AutoRenewalFailed` (promotion to real reservation is open follow-up).

#### Breaking changes

- `renew(block, index)` semantics change: was a synchronous renewal, now a one-shot scheduler. Clients expecting immediate renewal must migrate to `force_renew(TransactionRef::Position { ... })`.
- `renew_content_hash(content_hash)` removed. Use `force_renew(TransactionRef::ContentHash(_))` (sync) or `renew(TransactionRef::ContentHash(_))` (one-shot schedule).
- `Event::AutoRenewalEnabled` replaced with `Event::RenewalEnabled` (includes `recurring` field).


pallets/transaction-storage/src/lib.rs
```diff
@@ -881,6 +888,13 @@ pub mod pallet {
 				Error::<T>::AuthorizationNotFound,
 			);
 
+			// Consume one transaction unit from the caller's account authorization in
+			// lieu of a token fee. `size = 0` and `is_renew = false` keep the bytes and
+			// chain-wide `PermanentStorageUsed` counters untouched — actual bytes are
+			// charged on each renewal cycle by `do_process_auto_renewals`.
+			Self::check_authorization(&AuthorizationScope::Account(who.clone()), 0, true, false)
```
bkontur: @rosarp isn't this duplicate check then few lines above?
```
// Mirror `check_authorization`: the authorization must exist, be unexpired,
			// and have room for one more renewal of `tx_info.size`.
			let auth = Authorizations::<T>::get(AuthorizationScope::Account(who.clone()))
				.ok_or(Error::<T>::AuthorizationNotFound)?;
			ensure!(
				!Self::expired(auth.expiration) &&
					auth.extent.has_permanent_capacity(tx_info.size as u64),
				Error::<T>::AuthorizationNotFound,
			);
```

pallets/transaction-storage/src/lib.rs
```diff
@@ -1983,12 +1997,35 @@ pub mod pallet {
...
+						return Err(AUTHORIZATION_NOT_FOUND.into());
+					}
+					return Ok((
+						context.want_valid_transaction().then(|| ValidTransaction {
+							priority: T::StoreRenewPriority::get(),
+							longevity: T::StoreRenewLongevity::get(),
+							..Default::default()
+						}),
+						None,
+					));
```
bkontur, line 2171: @rosarp if we don't return here and let it flow, it will consume below `Self::check_authorization(`, then we don't need to charge in the extrinsic- we have pattern here, that we don't charge in extrinsic.

```suggestion
					(info.size as usize, info.content_hash, true)
```

pallets/transaction-storage/src/lib.rs
bkontur: @rosarp I would remove charging here and let the `check_signed` to charge as for other extrinsics: https://github.com/paritytech/polkadot-bulletin-chain/pull/504/files#r3204681541


And also maybe here, instead of `let who = ensure_signed(origin)?;`, we should to `ensure_authorized`.

rosarp: Yes — agreed, this duplicates the snapshot check. I'll move the validation + consumption into check_signed so the dispatch body becomes a clean.

karolk91:
I'm also thinking about making `disable_auto_renew` feeless - we would allow it when someone enabled it before - just for this account+item ? @bkontur what do you think? I'm only concerned if thats not a lot of logic to be done at the validation phase

karolk91:
I'm not convinced to this idea that enable_auto_renew just charges 1 tx - after some Claude conversations Im thinking about solution like this:


## `AccountActiveAutoRenewals` — reference

### What it is
Per-account denormalized view of `AutoRenewals`:
```rust
struct ActiveAutoRenewals { count: u32, total_bytes: u64 }
StorageMap<AccountId, ActiveAutoRenewals>
```
Invariant: always equals the count and byte-sum of `AutoRenewals` entries owned by that account.

### How it gates new registrations
At `enable_auto_renew` / manual renew validation — projection check, **no mutation**:
```
extent.transactions    + active.count       + 1     ≤ transactions_allowance
extent.bytes_permanent + active.total_bytes + size  ≤ bytes_allowance
```

### Scenarios

| Event | `extent` | `AutoRenewals` | `AccountActiveAutoRenewals` |
|---|---|---|---|
| `enable_auto_renew(h, S)` ok | — | `+ h` | `+(1, S)` |
| Duplicate / projection fails | — | — | — (rejected at pool entry, free) |
| Cycle success | `+1 tx, +size` | — | — |
| Cycle failure | — | `− h` | `−(1, S)` |
| `disable_auto_renew(h)` | — | `− h` | `−(1, S)` |
| Auth re-grant | reset to 0 | — | — (survives; projection picks up fresh extent) |
| `force_renew(h)` / manual renew | `+1 tx, +size` | — | — (projection must include `active` to avoid over-commit) |

Failure paths always converge on "remove from `AutoRenewals` + decrement summary," so the invariant is trivial to maintain and audit.

### Bottom line
"Register reserves, cycle pays." Costs one map + one MBM. Lookups are O(1), failure modes self-heal, enable/disable symmetric.

---

### How it relates to issue #517

#517 splits the API:
- `force_renew(hash)` — immediate manual renewal (today's `renew`).
- `renew(hash)` — *registered* one-shot, fires once at next retention boundary.
- `enable_auto_renew(hash)` — *registered* recurring, fires every cycle until disabled.

Both registered variants would gain a `recurring: bool` on `RenewalData`.

With the projection model, the one-shot vs recurring distinction reduces to **a single extra branch in the cycle-success path**:

```rust
if check_authorization(scope, size, consume=true, is_renew=true).ok() {
    do_renew_in_memory(...)
    if !renewal_data.recurring {
        AutoRenewals::remove(hash);
        AccountActiveAutoRenewals::mutate(owner, |a| { a.count -= 1; a.total_bytes -= size });
    }
}
```

Everything else is shared:
- Same projection check at register time (one-shot and recurring both reserve one cycle's worth).
- Same failure cleanup.
- Same disable path.
- `force_renew` validation also runs the projection so manual and registered paths compete on the same envelope honestly.


### Net
The projection model is the smaller mechanism that unifies #504's feeless goal and #517's one-shot/recurring split. Bundle #504 with the projection swap; #517 then lands as one extra branch.

rosarp:
> I'm also thinking about making `disable_auto_renew` feeless - we would allow it when someone enabled it before - just for this account+item ? @bkontur what do you think? I'm only concerned if thats not a lot of logic to be done at the validation phase

Will make it feeless on top of https://github.com/rosarp/polkadot-bulletin-chain/pull/8

franciscoaguirre:
@karolk91 That would make one single account the owner of the auto renewal. I thought we had agreed we should charge each time

pallets/transaction-storage/src/lib.rs
```diff
@@ -840,19 +840,26 @@ pub mod pallet {
...
-		/// renew-window quota (`bytes_permanent + tx_info.size <= bytes_allowance`) to
-		/// fit one more renewal. This mirrors the per-account hard cap that
-		/// `check_authorization` enforces at renewal time, on the correct axis
-		/// (`bytes_permanent`, not `bytes`). It does **not** guarantee the renewal will
-		/// succeed at process time — `bytes_permanent` may have advanced (other
-		/// renewals), the auth may have expired and been re-granted (counters reset),
-		/// or the chain-wide cap may have been hit. The actual gate is
+		/// Feeless: instead of charging a token fee, this call consumes **one transaction
+		/// unit** from the caller's account authorization at registration time. Bytes
+		/// are not charged here — they are charged on each auto-renewal cycle by
```
franciscoaguirre: By not charging bytes here. Someone can spam as many auto renewals as they want (up to `transactions_allowance`) when only one renew will succeed in the future. That would take away from the `MaxBlockTransactions` we have in the block when we could stop it if we charged. We could do a `force_renew` and schedule auto renewal so charging makes sense

pallets/transaction-storage/src/lib.rs
```diff
@@ -1212,9 +1209,25 @@ pub mod pallet {
...
+		/// - [`Self::check_authorization`] rejects — the auth was missing/expired, the per-account
+		///   renew quota (`bytes_permanent + size > bytes_allowance`) was exhausted, or the
+		///   chain-wide cap (`PermanentStorageUsed + size > MaxPermanentStorageSize`) would be
+		///   breached.
+		/// - [`Self::do_renew_in_memory`] returns `None` because the per-block transaction slot cap
+		///   (`MaxBlockTransactions`) is reached.
+		///
+		/// In both cases the on-chain data is left to expire normally on its existing
+		/// `RetentionPeriod` clock; the caller may re-`enable_auto_renew` once the
+		/// constraint clears.
```
franciscoaguirre: If auto-renewal fails, the data will expire in the next block, right? Then this comment is false

pallets/transaction-storage/src/lib.rs
```diff
@@ -1267,11 +1280,21 @@ pub mod pallet {
...
+		/// Called by:
+		/// - [`Self::do_renew`] for the single-renewal manual flow (`renew`, `renew_content_hash`).
+		/// - [`Self::do_process_auto_renewals`] in a loop, amortizing one [`BlockTransactions`]
+		///   read/write across all pending entries.
+		///
+		/// The hard-cap accounting (per-account `bytes_permanent`, chain-wide
+		/// [`PermanentStorageUsed`]) is performed by [`Self::check_authorization`] —
+		/// invoked by the extension's `pre_dispatch` for the manual flow and by
+		/// [`Self::do_process_auto_renewals`] for the auto flow before this is called.
+		fn do_renew_in_memory(
```
franciscoaguirre, line 1358: Why `in_memory`?

karolk91:
> @karolk91 That would make one single account the owner of the auto renewal. I thought we had agreed we should charge each time

What are you referring to ? Yes, currently only single account is the owner, and yes, I agree that we need to change that. But this PR seems orthogonal to the topic - multi account is something to be fixed by separate PR

pallets/transaction-storage/src/lib.rs
rosarp: the function operates on a &mut BoundedVec<...> accumulator passed in by the caller — it does NOT read or write the `BlockTransactions` storage slot itself.

`do_renew` reads/writes to `BlockTransactions`.

pallets/transaction-storage/src/lib.rs
rosarp: Updated the comment

pallets/transaction-storage/src/lib.rs
rosarp: I Agree. changed to force_renew + schedule auto renewal.

pallets/transaction-storage/src/lib.rs
rosarp: Now using `ensure_authorized`

franciscoaguirre:
@karolk91 I feel this adds a lot of complexity. We could instead charge on auto-renewal and toggle a flag on the renewal data that the renewal on the next period is already paid for. Then when we renew automatically we don't charge but we toggle the flag again. Following automatic renewals are going to be charged.

Disabling auto-renewal still works in this case but the user would've already been charged and wasted that first renewal. However, if we make it so the user can't disable it if this boolean flag is true, then we disallow cases where someone schedules auto-renewal only to not have enough allowance at the time of renewal and letting it drop.

I believe this solves the spam problem and also gives us an initial "grace period" where the account already paid for a first renewal. What do you think?

rosarp:
I agree above changes simplifies things & has lesser changes.

However with your suggested changes, I found 3 explicit caveats:
 - `bytes_permanent` semantics shift. Today it tracks renewals that actually added an on-chain entry. With this change, register-time charges show up in `bytes_permanent` without a matching `Renew` entry; the user's quota is consumed before any storage is reserved on-chain. 
  - `disable_auto_renew` becomes time-locked. A caller who registers and then immediately wants out has to wait until the first cycle fires (up to  `RetentionPeriod` blocks) before they can disable. That's the explicit UX cost.
  - Cycle-failure path drops a paid charge. If `PermanentStorageUsed` is at chain cap when the first cycle fires, the renewal fails and the  registration is cleaned up — but the user already paid `bytes_permanent` at register time. No refund mechanism exists today. (The @karolk91 suggested changes doesn't have this problem: register is free, so a failed cycle costs nothing.)

**Note**: double/multiple charge **does not** happen in current scenario as well as with below two models.

Comparison of two models : 

  | Surface | with `AccountActiveAutoRenewals` | `paid_for_next` only |
  |---|---|---|
  | New storage (version bump) | `AccountActiveAutoRenewals: AccountId → { count, total_bytes }` | None |
  | `check_signed` register (`renew` / `enable_auto_renew`) | `check_account_projection(who, size)` (read-only; sums `extent + active + size`) |  `check_authorization(scope, size, consume=true, is_renew=true)` (charges `bytes_permanent` + 1 tx slot; **no** `Transactions` entry added) |
  | `check_signed` `force_renew` / `renew_content_hash` | Standard  `check_authorization` **plus** prepended `check_account_projection` on the account branch | Unchanged (standard `check_authorization`) |
  | Register dispatch body | `AutoRenewals::insert(hash, RenewalData)` + `reserve_account_active(who,  size)` | `AutoRenewals::insert(hash, RenewalData { paid_for_next: true })` |
  | Cycle success (recurring) | `check_authorization(... true)` → `do_renew_in_memory` |  If `paid_for_next == true`: `do_renew_in_memory` + flip flag to `false` (no charge). Else: `check_authorization(... true)` → `do_renew_in_memory`   |
  | Cycle success (one-shot) | Same + `AutoRenewals::remove` + `release_account_active` | Same +  `AutoRenewals::remove` (flag inert at this point) |
  | Cycle failure | `AutoRenewals::remove` + `release_account_active`, emit `AutoRenewalFailed`  | `AutoRenewals::remove`, emit `AutoRenewalFailed` |
  | `disable_auto_renew` | Always allowed; decrements `AccountActiveAutoRenewals` | **Rejected while  `paid_for_next == true`** — caller must wait for first cycle |
  | `bytes_permanent` semantics | Bumped only at cycle | Bumped at register AND at every cycle except the "paid" first |
  | `PermanentStorageUsed` invariant | `Σ on-chain Renew sizes` (clean) | `Σ on-chain Renew sizes` (clean —  register doesn't add entries) |
  | Spam protection (bytes axis) | Hard: `extent + active + size ≤ bytes_allowance` |   Hard: `extent + size ≤ bytes_allowance` (because register itself mutates `extent`) |
  | Spam protection (tx axis) | Hard: `extent + active + 1 ≤ tx_allowance` | Soft unless you also add a tx-axis  hard check |
  | Double-charge risk | N/A (only cycle charges; active is just bookkeeping) | Prevented by the `paid_for_next` flag |
  | Cost of failed first cycle | None (no register-time charge) | **Paid `bytes_permanent` lost** — no refund  mechanism |
  | Auth refresh interaction | `extent` resets; `AccountActiveAutoRenewals` survives (reservations  carry over) | `extent` resets; `paid_for_next` survives → first cycle in new window is still free |

pallets/transaction-storage/src/lib.rs
```diff
@@ -722,59 +786,67 @@ pub mod pallet {
...
-			// and have room for one more renewal of `tx_info.size`.
-			let auth = Authorizations::<T>::get(AuthorizationScope::Account(who.clone()))
-				.ok_or(Error::<T>::AuthorizationNotFound)?;
-			ensure!(
-				!auth.expired(Self::now()) &&
-					auth.extent.has_permanent_capacity(tx_info.size as u64),
-				Error::<T>::AuthorizationNotFound,
-			);
+			let new_index = Self::do_renew(info)?;
+			Self::deposit_event(Event::Renewed { index: new_index, content_hash });
```
bkontur: @rosarp @karolk91 why does `enable_auto_renew` triggers immediately `renew` logic - transaction_index and so on? Shouldn't this be lazy?



```suggestion
```

pallets/transaction-storage/src/lib.rs
```diff
@@ -473,22 +492,67 @@ pub mod pallet {
...
 		/// ## Complexity
 		///
 		/// O(1).
-		#[pallet::call_index(1)]
-		#[pallet::weight((T::WeightInfo::renew(), DispatchClass::Operational))]
+		#[pallet::call_index(2)]
+		#[pallet::weight((T::WeightInfo::force_renew(), DispatchClass::Operational))]
 		#[pallet::feeless_if(|_origin: &OriginFor<T>, _block: &BlockNumberFor<T>, _index: &u32| -> bool { true })]
-		pub fn renew(
+		pub fn force_renew(
```
bkontur, line 549: @rosarp @karolk91 guys, please check this: https://github.com/rosarp/polkadot-bulletin-chain/pull/9

`renew_content_hash` should be also lazy as `fn renew`, I adjusted `force_renew` to handle both cases + some nits

rosarp:
@karolk91 @bkontur @franciscoaguirre 

projection model with AccountActiveAutoRenewals as per comment : https://github.com/paritytech/polkadot-bulletin-chain/pull/504#issuecomment-4422281633

https://github.com/rosarp/polkadot-bulletin-chain/pull/10

bkontur:
/cmd bench --runtime bulletin-paseo bulletin-westend --pallet pallet_bulletin_transaction_storage

bkontur:
/cmd bench --runtime bulletin-paseo bulletin-westend --pallet pallet_bulletin_transaction_storage

runtimes/bulletin-paseo/src/weights/pallet_bulletin_transaction_storage.rs
```diff
@@ -361,4 +361,11 @@ impl<T: frame_system::Config> pallet_bulletin_transaction_storage::WeightInfo fo
 			.saturating_add(T::DbWeight::get().reads(3))
 			.saturating_add(T::DbWeight::get().writes(2))
 	}
+	/// Placeholder weight for the v3→v4 stepped migration. Replace with a
+	/// benchmarked value when ready.
+	fn migrate_v3_to_v4_step() -> Weight {
```
bkontur, line 366: @rosarp why was this removed by bench bot?

bkontur:
/cmd bench --runtime bulletin-paseo bulletin-westend --pallet pallet_bulletin_transaction_storage

runtimes/bulletin-paseo/src/weights/pallet_bulletin_transaction_storage.rs
bkontur: @rosarp I see, we are missing benchmarking for `migrate_v3_to_v4_step`: https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/26094217617/job/76729914000?pr=504

bkontur (approved):

docs/001_authorizations.md
```diff
@@ -42,13 +49,13 @@ Each `TransactionInfo` is stamped with `kind: TransactionKind { Store, Renew }`.
...
 
 - **Soft (temporary)** — `bytes_allowance` and `transactions_allowance` are priority thresholds only. The boost stays on while in-budget on both axes (`bytes <= bytes_allowance` *and* `transactions <= transactions_allowance`) and drops to `0` once *either* is strictly over cap. A missing or `0`-allowance grant also yields no boost. `store` calls are never rejected; they queue behind in-budget signers when over.
-- **Hard (renewed)** — `bytes_allowance` is a real cap on the per-window renew quota. `renew` is **rejected** when `bytes_permanent + size > bytes_allowance`. The transaction-count axis does not gate renew. A separate chain-wide cap (`MaxPermanentStorageSize`) bounds the total renewed bytes on chain across all signers.
+- **Hard (renewed)** — `bytes_allowance` is a real cap on the per-window renew quota. `force_renew`, the `renew` one-shot at registration, and recurring auto-renewal cycles reject with `PermanentAllowanceExceeded` when `bytes_permanent + size > bytes_allowance`. The transaction-count axis does not gate renew. A separate chain-wide cap (`MaxPermanentStorageSize`) bounds the total renewed bytes on chain across all signers.
```
franciscoaguirre, line 52: "The transaction-count axis does not gate renew" I think it should, else you could renew 1 byte blobs and it's the same issue as with high-priority store() transactions. What do you think @bkontur @karolk91 ?

docs/001_authorizations.md
```diff
@@ -69,7 +76,7 @@ Extends `expiration` by another `AuthorizationPeriod` and leaves all consumed co
 Implemented by the [`AllowanceBasedPriority`][ext] transaction extension via a runtime-selected `BoostStrategy`:
 
 - `check_authorization` saturates `bytes` and `transactions` upward and never rejects.
-- The boost only applies to **signed account-scoped `store` / `store_with_cid_config`**. `renew` and preimage-scoped stores get `0`.
+- The boost only applies to **signed account-scoped `store` / `store_with_cid_config`**. All renew flavours and preimage-scoped stores get `0`.
```
franciscoaguirre, line 79: I also think renewals should get a priority boost. @karolk91 @bkontur

docs/001_authorizations.md
```diff
@@ -82,9 +89,14 @@ In-budget `store` txs sort strictly above over-budget ones. Pool nonce / arrival
 
 The hard cap is enforced at two levels, and a renewal that would breach **either** is rejected.
 
+"Renew" here means any path that consumes the per-window renew quota: `force_renew`,
+`enable_auto_renew`'s force-renew step, and each auto-renewal cycle in
+`do_process_auto_renewals`. The one-shot `renew` scheduler does **not** consume this
+quota at registration time — only at cycle time (via `do_process_auto_renewals`).
```
franciscoaguirre, line 95: I argue it should. We should take a decision on this before merging this PR

docs/001_authorizations.md
```diff
@@ -187,7 +199,7 @@ A user across many accounts (Sybil-like) is bounded by the chain-wide cap (Examp
 
 ## Migration
 
-`STORAGE_VERSION = 3`. Migrations are only relevant for the Paseo/Westend testnets carrying pre-existing on-chain state forward; see the `pallet_bulletin_transaction_storage::migrations::{v1, v2, v3}` modules for the wiring.
+`STORAGE_VERSION = 4`. Migrations are only relevant for the Paseo/Westend testnets carrying pre-existing on-chain state forward; see the `pallet_bulletin_transaction_storage::migrations::{v1, v2, v3, v4}` modules for the wiring. The v4 step re-encodes each `AutoRenewals` entry from `{ account }` to `{ account, recurring }` (all pre-existing entries were written by the old `enable_auto_renew`, so they migrate as `recurring: true`).
```
franciscoaguirre, line 202: I would also add the `paid: bool` field to know if the next cycle is paid, works for both one-shot and recurring renewals

pallets/transaction-storage/src/lib.rs
```diff
@@ -691,102 +748,92 @@ pub mod pallet {
...
-		/// [`Renewed`](Event::Renewed) event.
-		///
-		/// This is a convenience alternative to [`renew`](Self::renew) that does not require
-		/// knowing the exact `(block_number, tx_index)` pair.
-		///
-		/// Emits [`Renewed`](Event::Renewed) when successful.
-		#[pallet::call_index(10)]
-		#[pallet::weight((T::WeightInfo::renew_content_hash(), DispatchClass::Operational))]
-		#[pallet::feeless_if(|_origin: &OriginFor<T>, _content_hash: &ContentHash| -> bool { true })]
-		pub fn renew_content_hash(
```
franciscoaguirre, line 705: This is a breaking change, we need to make sure everyone using this on testnets knows about this beforehand

runtimes/bulletin-paseo/src/weights/pallet_bulletin_transaction_storage.rs
rosarp: @bkontur 
bot action is deleting and committing this - https://github.com/paritytech/polkadot-bulletin-chain/pull/504/changes/1900762ce13d00145ecafffba15f98fbb222b17a#diff-c7aeff1838c0ef6f332030f455a730edb3396d362612a87bb429baf61f10c438L366

rosarp: Fixed, added bench for it!

rosarp:
/cmd bench --runtime bulletin-paseo bulletin-westend --pallet pallet_bulletin_transaction_storage

PR #505, opened 2026-05-07
Skip bytes_permanent on re-renewal using existing TransactionKind

mudigal:
https://github.com/paritytech/polkadot-bulletin-chain/issues/506

**Summary**

- Replace the AccountRenewals storage map and live-tracking machinery with a simple is_re_renewal: bool derived from the existing TransactionInfo.kind field. When the entry being renewed already has kind ==
- Renew, it was previously renewed — skip the bytes_permanent capacity check and increment to avoid double-counting across renewal cycles. This is essential for content kept alive permanently: without the skip, every RetentionPeriod re-renewal would exhaust the per-account quota even though no new distinct content was added.
- Make renew_content_hash feeless (consistent with renew which was already feeless)
- Fix enable_auto_renew snapshot check to use the renew-axis (bytes_permanent) instead of the store-axis (bytes/transactions)
- Refactor on_initialize obsolete-block cleanup to a single pass (avoids a redundant iteration and removes a .clone())

**Test plan**
- cargo test -p pallet-bulletin-transaction-storage — 92 tests pass
- cargo clippy --all-targets --all-features --workspace -- -D warnings — clean
- Verify re-renewal across multiple RetentionPeriod cycles on a local testnet


bkontur:
I don't think we need this, we decrement/reset bytes_permanent when authorization window is expired and when creating new one for the same account: https://github.com/paritytech/polkadot-bulletin-chain/blob/cb01db7e766a9dc20099a7b38e761bfdbd8b1724/pallets/transaction-storage/src/lib.rs#L1519-L1525

We had "live per-account bytes_permanent tracking", but we decided to remove it and simplify stuff here: https://github.com/paritytech/polkadot-bulletin-chain/pull/486

mudigal:
> I don't think we need this, we decrement/reset bytes_permanent when authorization window is expired and when creating new one for the same account:
> 
> https://github.com/paritytech/polkadot-bulletin-chain/blob/cb01db7e766a9dc20099a7b38e761bfdbd8b1724/pallets/transaction-storage/src/lib.rs#L1519-L1525
> 
> We had "live per-account bytes_permanent tracking", but we decided to remove it and simplify stuff here: #486

After an elaborate discussion about the slots. Due to the tight coupling between Poeple Chain slots and Bulletin Chain slots, the scenario doesnt even arise. As the problem is only within the Authorisation period. So reverted everything and updated the Authorisations design doc.

docs/001_authorizations.md
```diff
@@ -198,17 +227,55 @@ When `PermanentStorageNearCap` fires governance can either:
...
-- **Specify `process_auto_renewals` behavior on chain-wide cap rejection.** If `do_renew` rejects an auto-renewal because of `MaxPermanentStorageSize`, treat it the same as PR #313's "block full" path: remove the registration, emit `AutoRenewalFailed`, let the data expire normally.
-- **Drop the snapshot check in `enable_auto_renew`** (or replace with a real reservation). The current check (`extent.transactions > 0 && extent.bytes >= tx_info.size`) is misleading and the per-window quota framing makes it even less meaningful — it suggests "this will work" while making no guarantees beyond the current block.
-- **Reserve block-transaction slots for user txs.** `process_auto_renewals` is mandatory and pushes into the same `BlockTransactions` slot as user `store`/`renew`. Cap auto-renewals to a fraction of `MaxBlockTransactions` or partition the slot budget.
-- **Per-content dedup of re-renewals (nice-to-have).** On `renew(X)`, look up the previous `(block, idx)` for `X` via `TransactionByContentHash` and cancel its pending decrement — drops the per-content double-count when the same content is renewed in multiple consecutive windows.
```
rosarp, line 213: all these are being worked on as TODO. 
These are still needed to be done right @bkontur ?

rosarp (approved):

PR #515, opened 2026-05-10
Centralize renewal mechanics in do_renew_in_memory

rosarp:
Centralize renewal mechanics in a single helper `do_renew_in_memory`:
stamp `kind = Renew`, push onto `BlockTransactions`, call `transaction_index::renew`, update `TransactionByContentHash`. Returns `None` at the per-block `MaxBlockTransactions` cap.

All three entry points go through it:
- `renew` / `renew_content_hash` via `do_renew` (`BlockTransactions::try_mutate`).
- Auto-renewal drain via `do_process_auto_renewals` (`BlockTransactions::mutate` wrapping the whole loop, so the `BoundedVec` is SCALE-encoded once per block instead of n times — preserves the O(n) batching).

Hard-cap accounting (per-account `bytes_permanent`, chain-wide `PermanentStorageUsed`) is unchanged: `check_authorization` handles it on the extension's `pre_dispatch` for manual renewals, and inline in `do_process_auto_renewals` for auto.

`do_process_auto_renewals`'s failure path now uniformly drops the registration and emits `AutoRenewalFailed` whether the gate failed on expired/missing auth, per-account cap, chain-wide cap, or block-slot cap; the on-chain data expires normally on its existing `RetentionPeriod`. When a cycle that had already been charged (a `paid = true` prepayment, or the per-cycle charge applied for the failing recurring case) is rejected by the per-block slot cap, the new `refund_renewal_charge` helper reverses the chain-wide `PermanentStorageUsed` bump so the cap does not leak; the per-account `bytes_permanent` / `transactions` increments are intentionally left in place — slot-cap rejection at inherent time is pathological (the inherent runs before user txs and `len(pending) <= MaxBlockTransactions`), and refunding into the current `Authorizations` entry would silently apply across auth roll-overs.

Also fixes the `check_permanent_storage_accounting` `try_state` invariant: `renew` / `enable_auto_renew` charge `PermanentStorageUsed` up front in `check_signed`, but the matching `Renew` entry only lands in `Transactions` once the cycle fires at the next retention boundary. The invariant now reconciles `PermanentStorageUsed` against `Σ Renew sizes in Transactions + Σ paid AutoRenewals sizes`, so try_state passes during the prepayment window.

Adds an `AUTO_RENEWAL_ALREADY_ENABLED` rejection in the extension's `validate_signed` for `renew`, mirroring the existing `enable_auto_renew` guard — duplicate registrations are rejected at pool admission before any charge, instead of charging in `pre_dispatch` and then erroring in the dispatch body.

Lays groundwork for the [TODO](https://github.com/paritytech/polkadot-bulletin-chain/blob/main/docs/001_authorizations.md#todo) items in `docs/001_authorizations.md`.


pallets/transaction-storage/src/lib.rs
```diff
@@ -1219,9 +1219,26 @@ pub mod pallet {
 		/// Drain [`PendingAutoRenewals`], renewing each entry whose owner has sufficient
 		/// authorization. Returns the count drained.
 		///
-		/// Batches the [`BlockTransactions`] read/write across all `n` renewals. Calling
-		/// [`Self::do_renew`] in the loop would re-encode the full vec per iteration (O(n²)),
-		/// which a linear weight model underestimates by ~17% at saturation.
+		/// Batches the [`BlockTransactions`] read/write across all `n` renewals by threading
+		/// an in-memory accumulator through repeated [`Self::do_renew_in_memory`] calls.
+		/// A naive [`Self::do_renew`]-per-item loop would re-encode the full vec per
+		/// iteration (O(n²)), which a linear weight model underestimates by ~17% at
```
bkontur: @rosarp please, tell your assistant "minimalize and simplify added docs" and "reflect changes to docs"

pallets/transaction-storage/src/lib.rs
```diff
@@ -1410,21 +1437,24 @@ pub mod pallet {
...
-				info.chunk_root,
-				info.size,
-				info.content_hash,
-				info.hashing,
-				info.cid_codec,
-				extrinsic_index,
-				TransactionKind::Renew,
-			)?;
-			sp_io::transaction_index::renew(extrinsic_index, content_hash);
+			let mut transactions = <BlockTransactions<T>>::get();
```
bkontur: @rosarp instead of get/put, we should be able to use `<BlockTransactions<T>>::try_mutate(|transactions| {` pattern - do we have similar places?

bkontur: also looks "strange" that "do_renew_in_memory" modifies some "transactions.try_push(new_info).ok()?;`, why it cannot mutate directly `BlockTransactions<T>`?  `BlockTransactions<T>` are never stored in DB right, it is transient stuff, so the performance shouldnt degradate?

rosarp: get/put changed to try_mutate.

rosarp: You're correct that no disk write happens. `BlockTransactions` is cleared via take() in on_finalize, so its value never reaches the on-disk trie at   block commit. The overlay holds it in memory for the block's duration.  The cost being paid is SCALE encode/decode, not disk I/O.

Why an &mut accumulator avoids it. mutate(|t| { … }) decodes once when entering the closure and encodes once when leaving it. Inside the closure,   mutations against &mut BoundedVec are pure in-memory Rust — no codec, no storage layer. So we open the closure, run the whole loop with  do_renew_in_memory(t, …) inside it, and close. That's the whole purpose of do_renew_in_memory taking &mut BoundedVec: it's the in-memory shape  that both try_mutate (single call) and mutate (loop) hand to their closures, so the same helper serves both paths without forcing a per-call  codec round-trip.

bkontur (approved):

PR #516, opened 2026-05-10
Add grace period to AuthorizationExtent

bkontur:
Introduces a renew-only grace window after `expiration`, controlled by `Config::GracePeriod`. Authorization lifecycle becomes:

- now < expiration: active (store + renew)
- expiration <= now < grace_until: in grace (renew only)
- now >= grace_until: expired (both rejected, eligible for remove_expired_*)

Highlights:

- `Authorization` gains `grace_until`; helpers `Authorization::expired(now)` / `::past_grace(now)` replace the previous `Pallet::expired` / `::past_grace`.
- `check_authorization` rejects store-in-grace; `check_authorization_expired` and `remove_expired_authorization` wait for `grace_until`.
- v3→v4 migration tail-extends `Authorization` with `grace_until = expiration + GracePeriod`; entries already past v4 grace are dropped (and `authorization_removed` called to release provider refs).
- v1→v2 migration now writes a frozen `V2Authorization` via a `From` impl, so later field additions to `Authorization` are a compile error in v2 rather than a silent shape change.
- Westend / Paseo runtimes set `type GracePeriod = AuthorizationPeriod` (14 days, matching the RFC).

---

_Part of #629 — cross-repo roadmap tracking._


pallets/transaction-storage/src/lib.rs
```diff
@@ -181,13 +181,34 @@ pub type AuthorizedCallerFor<T> = AuthorizedCaller<<T as frame_system::Config>::
...
+/// - `now >= grace_until`: expired — both rejected; eligible for `remove_expired_*`.
 #[derive(Encode, Decode, scale_info::TypeInfo, MaxEncodedLen)]
 struct Authorization<BlockNumber> {
 	/// Extent of the authorization (number of transactions/bytes).
 	extent: AuthorizationExtent,
-	/// The block at which this authorization expires.
+	/// The block at which this authorization expires (start of grace).
 	expiration: BlockNumber,
+	/// The block at which the grace window ends. Always `>= expiration`.
+	grace_until: BlockNumber,
```
rosarp: grace_until may not be required. 

 - it adds to storage
 - adds db v3->v4 migration, which can otherwise be skipped
 - Compute cost of deriving it on read: one `Get::get()` and one `saturating_add`. The calculation is strictly cheaper than reading the extra 4 bytes from storage on every access.

Can we avoid this field? And compute it instead?
If we avoid then functionally, the PR could be reduced to:

  - keep Authorization { extent, expiration } (no new field, no v3→v4 migration)
  - replace auth.past_grace(now) with now >= auth.expiration.saturating_add(T::GracePeriod::get())
  - drop migrations::v4 and the STORAGE_VERSION bump entirely
  - keep all the lifecycle semantics in check_authorization / remove_expired_authorization

pallets/transaction-storage/src/lib.rs
```diff
@@ -1627,7 +1653,7 @@ pub mod pallet {
 		/// promoting blobs for an account that has spent all of its store/renew quota.
 		pub fn account_has_active_authorization(who: &T::AccountId) -> bool {
 			Authorizations::<T>::get(AuthorizationScope::Account(who.clone()))
-				.is_some_and(|a| !Self::expired(a.expiration))
+				.is_some_and(|a| !a.expired(Self::now()))
```
rosarp, line 1471: External caller for this is `hop-promotion::can_account_promote (pallets/hop-promotion/src/lib.rs:106)`

This means HOP promotions stop the moment grace begins, even though the account can still renew. 
Is this intentional? 
Or this needs to be `past_grace` instead?

pallets/transaction-storage/src/lib.rs
```diff
@@ -883,7 +908,7 @@ pub mod pallet {
 			let auth = Authorizations::<T>::get(AuthorizationScope::Account(who.clone()))
 				.ok_or(Error::<T>::AuthorizationNotFound)?;
 			ensure!(
-				!Self::expired(auth.expiration) &&
+				!auth.expired(Self::now()) &&
```
rosarp, line 726: Same question, in grace_period, we will not allow `enable_auto_renew` call right?
May be this could be documented.

bkontur:
@rosarp thank you for comment, I will check later, I will postpone this PR for now and wait for slots

PR #519, opened 2026-05-11
zombienet-sdk auto-renew e2e suite

karolk91:
## Summary

Adds a zombienet-sdk e2e suite for the auto-renewal feature and overhauls how the zombienet CI runs.

- **16 new e2e tests** covering the full auto-renewal lifecycle (cadence, quota exhaustion, authorization expiry, block-weight under worst-case load, eviction under block pruning, behaviour across node restarts with changed pruning args).
- **Single consolidated CI workflow** for the entire zombienet-sdk suite (auto-renew + sync), replacing the two separate workflows.
- **Reproducible binary fetch** that can pin polkadot-sdk to either a release tag or a custom commit, with on-disk caching that survives across runs and CI cache distribution that doesn't rebuild binaries 8× per CI run.

## Test suite

All tests live in `zombienet-sdk-tests/tests/auto_renew_storage.rs` and run via `cargo test` with the `zombie-auto-renew-tests` feature.

Tests are organised into **groups by chain configuration**, with each group surfacing as a CI matrix entry and a `just` recipe target.

### Test scenarios

| Test | What it covers |
|---|---|
| `parachain_auto_renew_test` | Baseline — store + `enable_auto_renew` + observe N successful renewals; data stays bitswap-fetchable across cycles. |
| `parachain_auto_renew_authorization_expires_mid_cycle_test` | Cycle 1 renews; Alice's authorization expiration was overridden to fall between cycles, so cycle 2 emits `AutoRenewalFailed` and removes from `AutoRenewals`. Then re-authorize → fresh entry → another cycle 1 renew works. |
| `parachain_auto_renew_quota_exhaustion_test` | Alice's `bytes_allowance` overridden to exactly 2× data size. Cycles 1, 2 succeed; cycle 3 hits `PERMANENT_ALLOWANCE_EXCEEDED` → `AutoRenewalFailed`. |
| `parachain_auto_renew_many_items_test` | `MaxBlockTransactions` (512) items stored in one block, all enabled for auto-renew. Observes 3 successful renewal cycles each draining 512 items via `apply_block_inherents`; asserts block weight stays within `BlockWeights::max_block`; cycle 4 exhausts the authorization → 512 × `AutoRenewalFailed`. |
| `parachain_auto_renew_many_items_worst_case_test` | Same 512-item scenario, each item signed by a distinct worker so renewals touch 512 distinct `Authorizations` storage keys (no cache hits across iterations) — exercises the declared worst-case PoV cost. |
| `parachain_auto_renew_with_concurrent_store_test` | Times a fresh `store(data2)` to land in the same block as the auto-renewal inherent for `data1`. Verifies the inherent + regular store coexist correctly and both items end up indexed. |
| `parachain_auto_renew_vs_no_renew_eviction_test` | Side-by-side: two items, one with auto-renew, one without. After `--blocks-pruning=15` evicts the store block, the renewed item is still served via bitswap, the un-renewed item returns `DONT_HAVE`. |
| `parachain_renew_twice_within_block_with_pruning_test` | Alice + Bob each submit a `renew` for the same data in the same block (two distinct signers avoid the pool's `provides((who, content_hash))` tag conflict). After both store-block and renew-block are pruned, bitswap returns `DONT_HAVE` (col11 refcount reached 0). |
| `parachain_on_initialize_cleanup_test` | Isolates the `Hooks::on_initialize` weight cost from `apply_block_inherents` drain cost. Stores N items, exits before renewal, observes `on_initialize` cleanup at the expiry block. |
| `parachain_on_initialize_no_renewals_weight_test` | N stores without auto-renewal — measures the `on_initialize` expiry-sweep weight where `PendingAutoRenewals` is never populated. |
| `parachain_check_proof_fails_under_pruning_test` | `--blocks-pruning < RetentionPeriod`: at `store_block + RP`, the inherent provider can't construct a storage proof (col11 already evicted), the inherent isn't emitted, on_finalize panics — chain halts as expected. |
| `parachain_auto_renew_under_pruning_chain_halts_test` | Same pruning trap, but with auto-renewal scheduled past the proof block. Same halt outcome. |
| `parachain_restart_archive_to_pruning_test` / `_pruning_increase_test` / `_pruning_decrease_test` | Cold-restart the collator with a new `--blocks-pruning` value; verify it accepts the change and continues producing. Three scenarios: archive→pruning, pruning-tighten, pruning-loosen. |
| `parachain_long_running_pruning_soak_test` | Long-running pruning soak — skipped by the CI `--skip`-list by default. |

### Groups, matrix, and harness sharing

| Group | Chain config | Tests | Sharing pattern |
|---|---|---|---|
| `archive` | archive mode (no `--blocks-pruning`), RP=10 | 4 (basic, many_items, quota_exhaustion, auth_expires_mid_cycle) | One shared zombienet network per group |
| `pruning` | `--blocks-pruning=15`, RP=10 | 3 (vs_no_renew, concurrent_store, renew_twice) | One shared zombienet network per group |
| `restart` | each test restarts the collator with different `--blocks-pruning` args | 3 (archive→pruning, pruning-increase, pruning-decrease) | One network per test (restart IS the test) |
| `mixed` | heterogeneous per-test config | 5 (check-proof, chain-halts, many-items-worst-case, on_init cleanup, on_init no-renewals weight) | One network per test |

Shared groups use a `tokio::sync::OnceCell<Arc<SharedHarness>>`: the first test to enter the group spawns the network and sets `RetentionPeriod` via `set_retention_period_finalized`; subsequent tests reuse the cached handle. Statics aren't dropped at end of the test binary, so spawned processes live until the binary exits.

Every test entry waits for `wait_for_finalized_quiescence(collator1, …)` (waits for `finalized >= best + 2`) so the chain's finalized state catches up before the next test reads any state — avoids stale-nonce / unauthorized-state races.

### Determinism notes baked into the suite

- **Canonical block lookup**: store/renew helpers (`authorize_and_store_data`, `submit_store_signed`, `submit_renew_pair`) return the block number from `TransactionByContentHash` read **at the inclusion-block hash**, not subxt's `tx_in_block.block_hash().number()`. The latter can be one off from the canonical key in `Transactions[N]` if subxt's view follows a short-lived fork — which would make per-block renewal-event assertions and pruning math wrong.
- **Authorization override**: tests that need a tight Alice authorization (so renewals exhaust at exactly cycle N+1) use `override_alice_authorization` — a sudo `set_storage` direct write that REPLACES the storage entry. The pallet's regular `authorize_account` is additive on the unexpired path, so it can't be used to shrink Alice's genesis 10 MiB capacity.
- **Finalized authorize for first-tx accounts**: `authorize_account_via_sudo_finalized` is used where the immediate next step is a signed extrinsic from a fresh account (Bob in `renew_twice`, the 512 workers in `worst_case`) — the pool's `validate_signed` reads finalized state and would otherwise reject the first tx.
- **Explicit "test finished" condition**: every test that calls `enable_auto_renew` ends either with an observed `AutoRenewalFailed × expected_count` (authorization sized to exactly the observation window, so cycle N+1 exhausts) or an explicit `disable_auto_renew`. Net effect on the shared `archive` / `pruning` harnesses: the chain is observably idle between tests.
- **Polling negative-bitswap assertions**: `expect_bitswap_dont_have` polls (~1 Hz, up to `timeout_secs`) so the asynchronous block pruning + col11 refcount-zero cleanup that follows finality has time to land.

### Unified workflow

The previous `.github/workflows/auto-renew-tests.yml` and `sync-tests.yml` were consolidated into a single `.github/workflows/zombienet-tests.yml` with three jobs:

- **`prepare-binaries`** — runs once per workflow run; fetches/builds polkadot binaries + generates the bulletin chain specs.
- **`zombienet-auto-renew-tests`** — always runs on PR push. Matrix: 2 runtimes × 4 groups = 8 legs.
- **`zombienet-sync-tests`** — gated behind the `zombienet-sync-tests` PR label or manual `workflow_dispatch`. 2 legs (one per runtime).

Triggers:

```yaml
on:
  workflow_dispatch:
  pull_request:
    types: [opened, synchronize, reopened, labeled]
  push:
    branches: [main]   # post-merge runs populate main-scoped cache for PRs to read
```

### Custom polkadot-sdk revisions

All binary-version pins live in `.github/env` and can be either a release tag *or* a 40-character commit hash:

```bash
POLKADOT_NODE_VERSION=afba6ccb0a75908f2181ed0e849ddf827c71c501   # commit-pinned source build
CHAIN_SPEC_BUILDER_VERSION=polkadot-stable2603                   # release tag
FRAME_OMNI_BENCHER_VERSION=polkadot-stable2603
TRY_RUNTIME_VERSION=v0.10.1
ZOMBIENET_VERSION=v1.3.138
```

`scripts/get_polkadot_binaries.sh` resolves the value:

- **Release tag** (e.g. `polkadot-stable2603`) → downloads the prebuilt release asset from GitHub Releases, SHA-256 verified against `*_SHA256` env vars in `.github/env`.
- **40-char commit hash** → clones polkadot-sdk into `.polkadot-binaries/_src/` (reused for subsequent commits), source-builds with the right cargo features (e.g. `--features fast-runtime` for polkadot so westend-dev WASM is available), then stages the resulting binaries.

Both paths produce binaries at a versioned path:

```
.polkadot-binaries/<group>/<version>/<platform>/<binary>
```

where `<group>` is `polkadot-node` / `chain-spec-builder` / `frame-omni-bencher` / `try-runtime` / `zombienet`, and `<platform>` is e.g. `linux-x86_64` or `darwin-aarch64`. The `<version>` segment is whatever the env var resolves to, so multiple versions coexist on disk without conflict.

Override per shell for local work:

```bash
POLKADOT_NODE_VERSION=polkadot-stable2603 just binaries-polkadot
# or by commit hash (source-built):
POLKADOT_NODE_VERSION=afba6ccb0a75908f2181ed0e849ddf827c71c501 just binaries-polkadot
```

### Caching strategy

Two caches in CI, both `actions/cache@v5`:

| Cache | Path | Key | Purpose |
|---|---|---|---|
| Polkadot binaries | `.polkadot-binaries/polkadot-node` + `.polkadot-binaries/chain-spec-builder` (~333 MB; **explicitly excludes** the 11 GB `_src/` source tree + cargo target) | `polkadot-binaries-${runner.os}-${POLKADOT_NODE_VERSION}-${CHAIN_SPEC_BUILDER_VERSION}` | Stays warm across every PR that doesn't bump those env vars. The `push: main` trigger ensures every merge re-populates main's cache scope, which PRs whose base is main can read for free. |
| Bulletin chain specs | `zombienet/bulletin-westend-spec.json`, `zombienet/bulletin-paseo-spec.json` | `bulletin-specs-${runner.os}-${github.sha}` | Per-commit (the spec embeds the bulletin runtime WASM). Eliminates the ~4-min `cargo build -p bulletin-*-runtime` from every matrix leg (8× → 1×). |

`prepare-binaries` populates both caches. Matrix legs use `actions/cache/restore@v5` with `fail-on-cache-miss: true` — guaranteed hit because `prepare-binaries` finished before the leg runs (`needs: prepare-binaries`). No `actions/upload-artifact` / `actions/download-artifact` in the path (previous artifact-based approach hit `Unable to download after 5 retries` flakes and was ~10× slower than cache restore for the same data).

The chain-spec helper scripts `scripts/create_bulletin_*_spec.sh` are idempotent: if the spec file already exists they skip the runtime build, so a cache-restored spec passes through with no rebuild. Set `FORCE_REBUILD_SPEC=1` to override.

### Local development

`justfile` recipes mirror what CI does:

```bash
just binaries-polkadot                          # fetch/build polkadot binaries (respects env)
just chain-spec westend                         # build bulletin runtime + emit chain spec
just test-zombienet-auto-renew westend archive  # run a single group
just test-zombienet-auto-renew westend          # run all auto-renew tests (skips long soak)
just test-zombienet-sync westend                # run the sync suite
```


.github/env
```diff
@@ -1,17 +1,16 @@
 CI_IMAGE=docker.io/paritytech/ci-unified:bullseye-1.93.0-2026-01-27-v202603042356
 RUST_STABLE_VERSION=1.88.0
 NODE_VERSION=22
-POLKADOT_SDK_VERSION=polkadot-stable2603
 KUBO_VERSION=v0.40.1
-ZOMBIENET_VERSION=v1.3.138
+POLKADOT_NODE_VERSION=afba6ccb0a75908f2181ed0e849ddf827c71c501
```
bkontur:
```suggestion
# Actually used Polkadot-SDK `hop-base` branch
POLKADOT_NODE_VERSION=afba6ccb0a75908f2181ed0e849ddf827c71c501
```

.github/workflows/integration-test.yml
```diff
@@ -1,28 +1,19 @@
...
-  # Triggers the workflow on push or pull request events but only for the master branch
   push:
     branches: [main]
   pull_request:
     branches: [main]
-
-  # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
-# Cancel previous runs
```
bkontur, line 14: @karolk91 why just here? why not to clean this everywhere?

.github/workflows/zombienet-tests.yml
```diff
@@ -0,0 +1,213 @@
+name: Zombienet SDK Tests
+
+on:
+  workflow_dispatch:
+  pull_request:
+    # `labeled` lets the sync suite kick in when its label is added mid-PR.
+    types: [opened, synchronize, reopened, labeled]
+  # Populate the binaries cache under main's scope so PRs can read it.
+  push:
+    branches: [main]
```
bkontur, line 7: @karolk91 why not like other CIs?
```suggestion
on:
  workflow_dispatch:
  push:
    branches: [main]
  pull_request:
    branches: [main]
```

.github/workflows/integration-test.yml
```diff
@@ -49,53 +42,20 @@ jobs:
...
-      # Download and extract binaries if cache missed
-      # NOTE: zombienet releases do not publish checksums
-      - name: Download Zombienet binaries
-        if: steps.zombienet-cache.outputs.cache-hit != 'true'
-        env:
-          GITHUB_TOKEN: ${{ github.token }}
+          path: .polkadot-binaries
+          key: polkadot-binaries-${{ runner.os }}-${{ env.POLKADOT_NODE_VERSION }}-${{ env.CHAIN_SPEC_BUILDER_VERSION }}-${{ env.ZOMBIENET_VERSION }}
+
+      - name: Populate cache
```
bkontur, line 57: missing `if: steps.cache-binaries.outputs.cache-hit != 'true'?

.github/workflows/integration-test.yml
```diff
@@ -49,53 +42,20 @@ jobs:
...
         with:
-          path: ${{ env.ZOMBIENET_BIN_DIR }}
-          key: zombienet-${{ env.ZOMBIENET_VERSION }}-binaries
-      # Download and extract binaries if cache missed
-      # NOTE: zombienet releases do not publish checksums
-      - name: Download Zombienet binaries
-        if: steps.zombienet-cache.outputs.cache-hit != 'true'
-        env:
-          GITHUB_TOKEN: ${{ github.token }}
+          path: .polkadot-binaries
```
bkontur: It is not obvious how and where is `.polkadot-binaries` populated? inside `just binaries-polkadot` and `scripts/get_polkadot_binaries.sh`?

I mean, do we need both `just` and `scripts/get_polkadot_binaries.sh`? Can we "just" simplify this setup?

bkontur:
@karolk91 unrelated, can we please remove `log = "0.4"` and change it to `tracing` for `zombienet-sdk-tests` ?

.github/env
karolk91: we can't use comments in this file - I remember it causes syntax issues (it was likely on GH actions)

.github/workflows/integration-test.yml
karolk91: will do a cleanup in other places as well

.github/workflows/zombienet-tests.yml
karolk91: `opened, synchronize, reopened` are the defaults when not specified - so this is adding only `labeled` compared to others. Its used specifically by zombienet-based tests here

karolk91:
> @karolk91 unrelated, can we please remove `log = "0.4"` and change it to `tracing` for `zombienet-sdk-tests` ?

nice catch, of course - will change to tracing

.github/workflows/integration-test.yml
karolk91: it wasn't necessary because `get_polkadot_binaries.sh` script is idempotent (will exit quickly if cached files are present) - but now that you flagged it I think its worth to add this

.github/workflows/integration-test.yml
karolk91: I will extract that `.polkadot-binaries` as env variable. And I think `just` recipe + `scripts/get_polkadot_binaries.sh` is good be separated - its quite some bash code in there and it wouldn't look right

.github/workflows/zombienet-tests.yml
bkontur: > `opened, synchronize, reopened` are the defaults when not specified - so this is adding only `labeled` compared to others. Its used specifically by zombienet-based tests here

I see you removed "pull_request:" , so we don't want to run this on every PR?

.github/workflows/check-runtime-migration.yml
```diff
@@ -81,13 +74,20 @@ jobs:
...
 
-      - name: Install try-runtime-cli
+      - name: Ensure `just` is available
         if: steps.cache-restore.outputs.cache-matched-key == ''
-        run: |
-          curl -fsL "https://github.com/paritytech/try-runtime-cli/releases/download/${TRY_RUNTIME_VERSION}/try-runtime-x86_64-unknown-linux-musl" -o try-runtime
-          echo "${TRY_RUNTIME_SHA256}  try-runtime" | sha256sum -c -
-          chmod +x try-runtime
-          sudo mv try-runtime /usr/local/bin/
+        run: command -v just || cargo install just --locked
```
bkontur: `curl` is much much faster than `cargo install just` and then `curl`,
why can we just trigger this directly `./scripts/get_polkadot_binaries.sh try-runtime` and skip unnecessary `just` installation?

bkontur: `just` installation took 1.5minute and in the end, it just triggers `curl`

image

.github/workflows/check.yml
```diff
@@ -178,27 +170,17 @@ jobs:
       - name: Load common environment variables via env file
         run: cat .github/env >> $GITHUB_ENV
 
-      - name: Create directory for binary and add to PATH
-        run: |
-          mkdir -p $FRAME_OMNI_BENCHER_BIN_DIR
-          echo "${FRAME_OMNI_BENCHER_BIN_DIR}" >> $GITHUB_PATH
+      - name: Ensure `just` is available
+        run: command -v just || cargo install just --locked
```
bkontur: the same, wasted 1.5 minute for installing `just`, why not directly `./scripts/get_polkadot_binaries.sh frame-omni-bencher` ?

```suggestion
        run: command -v just || cargo install just --locked
```

.github/workflows/zombienet-tests.yml
karolk91: `pull_request:` is still there - not removed

zombienet-sdk-tests/tests/utils/sync.rs
```diff
@@ -133,6 +133,22 @@ pub async fn wait_for_finalized_height(
 	.context(format!("Node did not finalize block height {}", min_height))
 }
 
+/// Wait until finality catches up to (current best + 2). `account_nonce()` reads finalized
+/// state, so without quiescence the next test sees a stale nonce.
+pub async fn wait_for_finalized_quiescence(
```
bkontur, line 138: I've never seen this word "quiescence" :D :D :D

.github/workflows/check.yml
karolk91: My general idea was that we should be able to easily run tests also locally on PC - so using `just` in both cases (local + CI) did make sense to me

.github/workflows/check-runtime-migration.yml
karolk91: I think it would make sense to have `just` preinstalled in the docker image already - I can propose a change to the Dockerfile. I was also thinking about this before and decided to use cargo, but I can change to binary download instead

examples/justfile
```diff
@@ -162,10 +167,10 @@ bulletin-parachain-zombienet-start test_dir runtime="bulletin-westend-runtime":
         exit 1
     fi
 
-    # Create chain spec
+    # Create chain spec — make sure chain-spec-builder is on PATH.
     echo "   Creating chain spec for {{ runtime }}..."
-    ROOT_DIR="{{ justfile_directory() }}/.."
-    PATH="$HOME/local_bulletin_testing/bin:$PATH" "$ROOT_DIR/scripts/$SPEC_SCRIPT"
+    CSB_DIR="$("$ROOT_DIR/scripts/get_polkadot_binaries.sh" chain-spec-builder)"
```
bkontur: few lines above we are doing "just binaries-chain-spec-builder", why not here?

.github/workflows/check-runtime-migration.yml
bkontur: yeah, that would be cool to have `just` preinstalled, but, ok, if we want to use `just` then let's make sure, it is cached one, so we don't install it, we can add just version to env and build it once or something like that

.github/workflows/check.yml
bkontur: > My general idea was that we should be able to easily run tests also locally on PC - so using `just` in both cases (local + CI) did make sense to me

ok, but let's make sure we do caching in CI properly.

Before the pattern was simple, if cache was not hit, it downloaded or build, now I am little bit confused, if the caching is handled on CI level or `get_polkadot_binaries.sh`?
```
- name: Cache frame-omni-bencher

- name: Download frame-omni-bencher
        if: steps.frame-omni-bencher-cache.outputs.cache-hit != 'true'
```

.github/workflows/zombienet-tests.yml
bkontur: > `pull_request:` is still there - not removed

I see, a morning blindness without the coffee :D

.github/workflows/check-runtime-migration.yml
bkontur: would it make sense to cover all the just installation cache "- name: Ensure `just` is available" to separate file similar like we have check-changed-files.yml?

bkontur (approved):

scripts/get_polkadot_binaries.sh
```diff
@@ -0,0 +1,272 @@
...
+			if [ -n "$llvm_prefix" ] && [ -d "$llvm_prefix/lib" ]; then
+				export DYLD_FALLBACK_LIBRARY_PATH="$llvm_prefix/lib${DYLD_FALLBACK_LIBRARY_PATH:+:$DYLD_FALLBACK_LIBRARY_PATH}"
+			fi
+		fi
+		targets="$(build_targets_for_group "$GROUP")"
+		if [ "$GROUP" = "polkadot-node" ]; then
+			# `--features fast-runtime` enables zombienet's `westend-development` preset
+			# with short epochs; not a feature of polkadot-omni-node, so split builds.
+			log "  cargo build --release --locked -p polkadot --features fast-runtime"
+			cargo build --release --locked -p polkadot --features fast-runtime
```
bkontur, line 243: @karolk91 teoretically, we don't need to build polkadot* binaries, we can just download them from release, https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/25860302346/job/75988362206?pr=519#step:7:73, this could save most of the CI time - at least it could be configurable from `env` (build or download).

I got the situation around polkadot-omni-node, because of `hop-base` branch, but it should also be fixed by configuration from `env`

bkontur:
@karolk91 looks like one failed: https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/25860302346/job/75991988761

scripts/get_polkadot_binaries.sh
karolk91: current implementation is that it will build the binary when its some commit hash provided - otherwise it will try to download

PR #523, opened 2026-05-12
Add Bulletin chainspecs for Polkadot, Westend, and Paseo

mudigal:
## Summary
- Add `chainspecs/` directory with Bulletin chain specifications copied from `paritytech/chainspecs`
- Includes Polkadot, Westend, and Paseo networks
- Each network has `chainspec.json` and `raw.json` (Polkadot also includes `genesis-head`)

## Structure
```
chainspecs/
├── paseo/
│   ├── chainspec.json
├── polkadot/
│   ├── chainspec.json
└── westend/
    ├── chainspec.json
```

## Enables 
- Merging of https://github.com/paritytech/chainspecs/pull/1263

## Closes
- Partially addresses https://github.com/paritytech/polkadot-bulletin-chain/issues/499

PR #537, opened 2026-05-14
ci: chainspec sync check

x3c41a:
## Summary
- Add CI workflow that boots `polkadot-omni-node` for each chainspec in `chainspecs/` and waits until it imports a few blocks
- `polkadot-omni-node` version (and SHA256) is pinned in the workflow itself
- Matrix runs Polkadot, Westend, Paseo Next, and Paseo Next v2 in parallel, each against a public relay-chain RPC

## Triggers
- Pushes/PRs that touch `chainspecs/` or the workflow itself
- `workflow_dispatch`

## Follow-up to
- #523


.github/workflows/chainspec-sync-check.yml
```diff
@@ -0,0 +1,140 @@
...
+
+      - name: Download polkadot-omni-node
+        if: steps.omni-node-cache.outputs.cache-hit != 'true'
+        run: |
+          mkdir -p "$OMNI_NODE_BIN_DIR"
+          cd "$OMNI_NODE_BIN_DIR"
+          curl -fL -o polkadot-omni-node \
+            "https://github.com/paritytech/polkadot-sdk/releases/download/${OMNI_NODE_VERSION}/polkadot-omni-node"
+          echo "${OMNI_NODE_SHA256}  polkadot-omni-node" | sha256sum -c -
+          chmod +x polkadot-omni-node
```
bkontur: Let's use the same download mechanism: https://github.com/paritytech/polkadot-bulletin-chain/blob/273b188378a5358ed26b30a0e531eec4716c7471/.github/workflows/integration-test.yml#L129-L146

@karolk91  not sure if just `${{ env.POLKADOT_BINARIES_DIR }}/polkadot-omni-node` works

```suggestion     
```

.github/workflows/chainspec-sync-check.yml
```diff
@@ -0,0 +1,141 @@
...
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+env:
+  POLKADOT_NODE_VERSION: polkadot-stable2603
+  POLKADOT_OMNI_NODE_SHA256: a96cb007f3e32de3b0d77dc25cab324d7067cca26f1a5aad4d05d748cf89e323
+  POLKADOT_BINARIES_DIR: ${{ github.workspace }}/.polkadot-binaries
```
karolk91: any reason to override these ?
```suggestion
```

.github/workflows/chainspec-sync-check.yml
x3c41a: done

.github/workflows/chainspec-sync-check.yml
x3c41a: My bad, I did not realise that `.github/env` was by design

x3c41a: reverted

.github/workflows/chainspec-sync-check.yml
```diff
@@ -0,0 +1,146 @@
...
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - name: Load common environment variables via env file
+        run: cat .github/env >> $GITHUB_ENV
+
+      - uses: ./.github/actions/install-just
+
+      - name: Cache polkadot-node binaries
```
bkontur: @x3c41a @karolk91 I think we need to run "setup" before matrix job, because all took 16min:
https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/25871727728/job/76028482209?pr=537#step:6:32
https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/25871727728/job/76028482294?pr=537#step:6:32

`Cache not found for input keys: polkadot-node-Linux-d6a4f5977b39bf5e5152e2f2bb6719ea92b992ea`

karolk91: Yes, I was also looking at this. Here is some attempt to make it easier to handle these binaries+caches: #540. If we merge that then this job would need to do something like:

1. Outside of matrix job - to be run only by single runner, so if we need to recompile for some reason, then it will be only done on a single runner
```
      - name: Set up Polkadot binaries
        uses: ./.github/actions/use-polkadot-binaries
        with:
          groups: polkadot-node
          mode: prepare
```
2. Then inside a matrix job:
```
      - name: Set up Polkadot binaries
        uses: ./.github/actions/use-polkadot-binaries
        with:
          groups: polkadot-node
          mode: consume
```

.github/workflows/chainspec-sync-check.yml
```diff
@@ -0,0 +1,170 @@
...
+          POLKADOT_DIR="$(just binaries-polkadot)"
+          echo "$POLKADOT_DIR" >> "$GITHUB_PATH"
+
+      - name: Sync a few blocks
+        shell: bash
+        env:
+          CHAINSPEC: ${{ matrix.chainspec }}
+          RELAY_RPC: ${{ matrix.relay_rpc }}
+        run: |
+          set -euo pipefail
```
karolk91: Please extract this logic to a separate script

x3c41a: extracted

karolk91 (approved):

Issue #542, opened 2026-05-19
[Disable] Disable all *renew* extrinsic on demand


bkontur:
- runtime configuration (with `pub storage` const that can be override by OpenGov)
- hard-coded versions - tx extension or CallFilter?

TODO: find some good approach - preffered dynamic way

`#Bulletin-strike`

bkontur:
Potentially, this could be solved just by @rosarp's splitting the pallets: https://github.com/paritytech/polkadot-bulletin-chain/pull/571

PR #545, opened 2026-05-19
Disable renew calls on demand

RafalMirowski1:
https://github.com/paritytech/polkadot-bulletin-chain/issues/542

⏺ Disable renew calls on demand via pallet-tx-pause

  Integrates pallet-tx-pause into both bulletin-westend and bulletin-paseo runtimes as a dynamic kill-switch scoped to the renew family of pallet-bulletin-transaction-storage.

  What this enables

  - Halt renewals without a runtime upgrade. Root can call
  txPause.pause(("TransactionStorage", "renew")) and the call is rejected at dispatch with frame_system::Error::CallFiltered 
  - Per-call granularity. Each of renew, renew_content_hash, enable_auto_renew, disable_auto_renew can be paused independently.
  - Reversible. txPause.unpause(...) flips it back.
  - inverted whitelist (bulletin_pallets_common::TxPauseWhitelist<Runtime, TransactionStorage>) makes only the four renew calls pausable; any attempt to pause anything else returns pallet_tx_pause::Error::Unpausable.

  What stays unaffected 

  - apply_block_inherents is an inherent and bypasses BaseCallFilter — already-enabled auto-renewals keep flowing through the inherent while the manual + management extrinsics are paused.
  - sudo(renew) uses dispatch_bypass_filter, so root can still trigger renews while the pause is active.


pallets/common/src/lib.rs
```diff
@@ -408,6 +408,42 @@ pub fn proxy_inner_calls<T: pallet_proxy::Config>(
...
+		{
+			let _ = name;
+			false
+		}
+		#[cfg(not(feature = "runtime-benchmarks"))]
+		{
+			let is_renew_call = name.0.as_slice() == StoragePallet::name().as_bytes() &&
+				matches!(
+					name.1.as_slice(),
+					b"renew" | b"force_renew" | b"enable_auto_renew" | b"disable_auto_renew"
```
bkontur: this is not nice and it is error prone, doesnt pallet_tx_pause support better filtering?

RafalMirowski1: good point, is this better https://github.com/paritytech/polkadot-bulletin-chain/pull/545/changes/3339e2760e69834a38ffeadcb178a721447e1327 ?

bkontur:
@ggwpez if you have some time, could you please also check this usage of tx-pallet?

PR #548, opened 2026-05-20
Auto renew paid flag

franciscoaguirre:
One-shot `renew` pre-pays for the first and only cycle. I made it so `enable_auto_renew` also pre-pays for the first cycle. After that it goes back to charging every cycle. Users can't disable auto renewed when it's prepaid so people can be sure that that data will be renewed.

### DIFF
--- a/pallets/transaction-storage/src/lib.rs
+++ b/pallets/transaction-storage/src/lib.rs
@@ -1373,14 +1373,15 @@ pub mod pallet {
 						// Recurring: consume the prepayment so subsequent cycles
 						// charge per-cycle, and unblock `disable_auto_renew` for the
 						// owner now that the prepaid renewal has been delivered.
-						AutoRenewals::<T>::insert(
-							content_hash,
-							RenewalData {
-								account: renewal_data.account.clone(),
-								recurring: true,
-								paid: false,
-							},
-						);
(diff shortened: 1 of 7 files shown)


pallets/transaction-storage/src/lib.rs
```diff
@@ -1305,23 +1350,37 @@ pub mod pallet {
...
+						// charge per-cycle, and unblock `disable_auto_renew` for the
+						// owner now that the prepaid renewal has been delivered.
+						AutoRenewals::<T>::insert(
+							content_hash,
+							RenewalData {
+								account: renewal_data.account.clone(),
+								recurring: true,
+								paid: false,
+							},
+						);
```
bkontur: @franciscoaguirre let's not re-insert, just mutate to false, in case root disable auto renewal

```suggestion
						// `mutate` (not `insert`) so a Root `disable_auto_renew`
						// executed earlier in the same block — between the
						// `on_initialize` queue and this inherent — is not silently
						// re-armed by a fresh insert.
						AutoRenewals::<T>::mutate(content_hash, |entry| {
							if let Some(data) = entry {
								data.paid = false;
							}
						});
```

bkontur (approved):

PR #555, opened 2026-05-21
[WIP] make SDK great again

karolk91:
Reworks both off-chain client SDKs (TypeScript and Rust) into one submission engine that stores data reliably and at bulk throughput, and brings every example, the console UI, and CI onto it. No runtime/pallet changes.

## Why

The previous clients signed and watched one transaction at a time and trusted `signSubmitAndWatch` to carry it to finality. That breaks in the three situations the Bulletin chain actually runs in:

- **Bulk uploads.** A large file is many `store` extrinsics; one-at-a-time submission wastes block space and never fills the chain's per-block capacity.
- **Reorgs / fork churn.** A nonce read from a best block that later reorgs, or an  era anchored to a dropped block, silently invalidates in-flight transactions — surfacing as "bad nonce" / "usurped" / stalls.
- **Paying twice.** On any retry (reconnect, watchdog, dropped tx) a naive client re-signs with a fresh nonce and can re-submit the same content. For a paid store that is a correctness bug: **a session must upload each CID exactly once, even across terminal errors and reconnects.**

The light-client (smoldot) path made all three worse and was effectively unusable.

## The engine (shared design, idiomatic in each language)

- **Wave batching** sized to the runtime's weight/length budget and `MaxBlockTransactions`, so a bulk upload fills blocks instead of trickling.
- **`system_accountNextIndex` as the nonce source of truth** — reflects included *and* pool-pending txs and drops on reorg, so nonce assignment tracks reality.
- **Mortality anchored at a finalized block** — cannot reorg out, so txs stay valid across fork churn without re-signing.
- **Exactly-once via per-item nonce + TBCH.** Inclusion is confirmed by reading `TransactionByContentHash` at the finalized block (fork-walking), not a transient event. Each item owns a nonce slot that persists across retries; re-broadcast replaces the slot, never double-pays. A slot taken by another signer is detected and surfaced, not silently double-spent.
- **Watchdogs + retry-resume.** A chainHead stall / no-progress / disconnect triggers a re-subscribe and re-broadcast of not-yet-confirmed items at their carried nonce (still exactly-once); only after the retry budget does it surface a terminal error.

## Provider abstraction

The client is **agnostic to how it reaches the chain** — it takes *providers*, not URLs:

- **TypeScript:** `providers: () => [getWsProvider(url) | getSmProvider(spec)]`.
- **Rust:** `TransactionClient::from_rpc_clients(Vec<RpcClient>)` — a WS connection or a smoldot light client are both just an `RpcClient`.

The first provider is the monitor (reconcile + nonce/storage); **every** provider is a broadcast target (accepted-if-any), so one dead provider can't stall a run. The SDK owns the connection lifecycle (TS `destroy()`; Rust via `Drop`).

## Client surface

- **One upload primitive:** `estimateUpload(items | source)` → `submit(estimate, source)`. The estimate previews cost and dedups (within-input + optional on-chain `skipExisting`); `submit` drives to the requested confirmation level.
- **Streaming.** Upload a file from a re-openable byte source, range-read lazily, so resident memory tracks the in-flight window, not the whole file.
- **Unsigned (preimage) path** for bare extrinsics with no signer.
- **Batch / sudo authorize** (`Utility.batch_all`, `Sudo.sudo`) and separate upload vs authorizer signers.
- **Per-item events** keyed by index + CID (`Started → InBlock → Finalized`).

## Renewal & runtime compat

- `renew` / `forceRenew` take a `(block, index)` position or a content hash (`TransactionRef`), in both SDKs.
- The fleet runs several runtime generations at once, so renew is dispatched through a compat registry: the connected chain's metadata checksum selects the encoder (committed snapshots in `sdk/metadata-compat/`), failing closed on unknown shapes. Legacy positional runtimes still renew by position; content-hash renew / `force_renew` there error out cleanly.
- Both SDKs bump to 1.0.0.

## Rust ↔ TypeScript parity

The Rust SDK now mirrors the TS engine — same guarantees, events, and core error codes — implemented idiomatically (config structs over fluent builders; `Drop` over `destroy()`). Public surfaces match: `estimate_upload`/`submit`, `SeekableSource`, the provider abstraction, unsigned, batched authorize.

## Examples, console UI, tests, CI

- Every example runs on the SDK over both **ws and smoldot** via one shared provider helper; the console UI uploads through the same path.
- Tests cover the new failure modes directly: exactly-once accounting (same- and cross-account), hijack recovery, the unsigned path, streaming, dedup/guard, multi-provider fan-out.
- **CI** runs the TS vitest integration suite and the Rust `pipeline_live` suite against a node-backed zombienet (`runtime × {ws, smoldot}` matrix), gating the light-client path. SDK book docs rewritten for the new surface.

## Not in scope

- Runtime / pallet changes (none).
- Data retrieval — still out of band; unchanged.


bkontur:
@karolk91 if you go with sdk to bulletin-deploy, maybe worth to check if we could replace https://github.com/paritytech/bulletin-deploy/pull/693 with sdk

PR #563, opened 2026-05-26
Cleanup AllowedAuthorizers origin handling

franciscoaguirre:
Changed `type Authorizer: EnsureOrigin<Self::RuntimeOrigin>` to return an `Option<AuthorizationOrigin>` on success. This new struct represents allowed authorizers that we want to take budget from whenever they authorize. This change makes the pallet code simpler by using FRAME's built-in origin system


pallets/transaction-storage/src/lib.rs
```diff
@@ -168,8 +168,17 @@ pub mod pallet {
 		type AuthorizationPeriod: Get<BlockNumberFor<Self>>;
 		/// The origin that manages the authorizer list.
 		type AuthorizerRegistrarOrigin: EnsureOrigin<Self::RuntimeOrigin>;
-		/// The origin that can authorize data storage.
-		type Authorizer: EnsureOrigin<Self::RuntimeOrigin>;
+		/// The origin that can authorize data storage. `Success` is
+		/// `Some(AuthorizationOrigin { .. })` when the dispatcher is an
+		/// [`AllowedAuthorizers`] entry — carrying the budget owner, the
+		/// per-authorizer `authorization_period` override, and the authorizer's
+		/// `valid_until` (exposed for inspection; not used to clamp the granted
```
franciscoaguirre: While cleaning up I found that `valid_until` doesn't clamp the authorization. Wouldn't it be better to remove authorization_period_override and just clamp authorizations created via an AllowedAuthorizer to valid_until? @bkontur @karolk91 

Example:
There's a hackathon and governance allows a particular contract as an authorizer. We set `valid_until` for the end date of the hackathon. Then all authorizations given by this authorizer will be max until that time. I don't see the authorization period override providing any value in this case.

pallets/transaction-storage/src/lib.rs
```diff
@@ -992,15 +988,13 @@ pub mod pallet {
...
-			AllowedAuthorizers::<T>::try_mutate_exists(&who, |maybe_budget| {
-				let budget = maybe_budget.as_ref().ok_or(Error::<T>::AuthorizerNotFound)?;
-				ensure!(
-					Self::authorizer_removable(budget),
-					Error::<T>::AuthorizerBudgetNotExhausted,
-				);
-				*maybe_budget = None;
-				Ok::<_, DispatchError>(())
-			})?;
+			let budget =
```
bkontur, line 984: just asking, why get+remove vs try_mutate_exists?

pallets/transaction-storage/src/lib.rs
```diff
@@ -622,22 +631,14 @@ pub mod pallet {
 			transactions: u32,
 			bytes: u64,
 		) -> DispatchResult {
-			let authorization_period_override = Self::authorization_period_override_for(&origin);
-			// Capture the signer (if any) before `ensure_origin` consumes the origin —
-			// we use it post-check to charge their `AllowedAuthorizers` budget. Root /
-			// XCM origins have no signer and no budget to consume.
-			let signer = frame_system::ensure_signed(origin.clone()).ok();
-			T::Authorizer::ensure_origin(origin)?;
+			let signer = T::Authorizer::ensure_origin(origin)?;
```
bkontur: very nit I would rename signer -> authorizer or maybe_authorizer here

```suggestion
			let maybe_authorizer = T::Authorizer::ensure_origin(origin)?;
```

pallets/transaction-storage/src/types.rs
```diff
@@ -282,6 +282,31 @@ impl<BlockNumber: PartialOrd + Copy> AuthorizerBudget<BlockNumber> {
...
+///   downstream logic. The pallet does **not** use this to cap the granted
+///   authorization — authorizations are independent of their grantor's
+///   remaining validity.
+///
+/// `None` (as the full [`EnsureOrigin::Success`]) means the dispatcher is a
+/// non-account authorizer (Root / XCM / signed-by list) — no budget to charge
+/// and no overrides.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct AuthorizationOrigin<AccountId, BlockNumber> {
+	pub signer: AccountId,
```
bkontur:
```suggestion
	pub authorizer: AccountId,
```

pallets/transaction-storage/src/lib.rs
franciscoaguirre: Looks better in my opinion. I did it while I was at it but I can revert.

pallets/transaction-storage/src/lib.rs
bkontur: Hmm, good point, we could just validate or adjust: now + period_override` < `valid_until`)`. I added `period_override`as a completeness feature, to have better control, just in case to cover everything, I don't remember the feature/use-case I had in my mind that time.

runtimes/bulletin-paseo/src/storage.rs
```diff
@@ -119,16 +119,28 @@ impl pallet_bulletin_transaction_storage::Config for Runtime {
...
-				EnsureXcm<IsSiblingParachain>,
+				EnsureAnonymousAuthorizer<
+					EnsureXcm<IsSiblingParachain>,
+					Self::AccountId,
+					crate::BlockNumber,
+				>,
 			>,
 			// Test accounts can also authorize for testing purposes.
-			EnsureSignedBy<TestAccounts, Self::AccountId>,
+			EnsureAnonymousAuthorizer<
```
bkontur: I am not sure about the name `EnsureAnonymousAuthorizer`, why `Anonymous`? especially here? We are using exact `TestAccounts` a.k.a. known list of accounts.
```
/// non-account-bearing authorizers (Root, XCM, signed-by lists)
```

pallets/transaction-storage/src/types.rs
```diff
@@ -324,3 +352,28 @@ where
...
 	}
 }
+
+/// `EnsureOrigin` adapter that wraps an inner origin and projects its `Success` to
+/// `None: Option<AuthorizationOrigin<AccountId, BlockNumber>>`. Used to align
+/// non-account-bearing authorizers (Root, XCM, signed-by lists) with the
+/// `Option<AuthorizationOrigin<_, _>>` `Success` produced by
+/// [`EnsureAllowedAuthorizers`] when composing the [`Config::Authorizer`] chain
+/// via [`EitherOf`](polkadot_sdk_frame::deps::frame_support::traits::EitherOf).
+pub struct EnsureAnonymousAuthorizer<E, AccountId, BlockNumber>(
```
bkontur: can we just remove `Anonymous`? :)
```suggestion
pub struct EnsureAuthorizer<E, AccountId, BlockNumber>(
```

pallets/transaction-storage/src/types.rs
```diff
@@ -324,3 +352,28 @@ where
 		Ok(frame_system::RawOrigin::Signed(who).into())
 	}
 }
+
+/// `EnsureOrigin` adapter that wraps an inner origin and projects its `Success` to
+/// `None: Option<AuthorizationOrigin<AccountId, BlockNumber>>`. Used to align
+/// non-account-bearing authorizers (Root, XCM, signed-by lists) with the
+/// `Option<AuthorizationOrigin<_, _>>` `Success` produced by
+/// [`EnsureAllowedAuthorizers`] when composing the [`Config::Authorizer`] chain
+/// via [`EitherOf`](polkadot_sdk_frame::deps::frame_support::traits::EitherOf).
```
bkontur: I would not suggest using `EitherOf`, I mean `EnsureAnonymousAuthorizer` can be used with different stuff `EitherOrDiverse` and other helpers also.

```suggestion
/// `Option<AuthorizationOrigin<_, _>>`.
```

bkontur (commented):
@franciscoaguirre I like this PR and changes, left just some naming nits. Just thinking if we could also simplify/reuse `EnsureAnonymousAuthorizer/EnsureAuthorizer` for `EnsureAllowedAuthorizers` (maybe not needed). I will do a review morning

franciscoaguirre:
@bkontur Addressed feedback

runtimes/bulletin-paseo/src/storage.rs
```diff
@@ -100,12 +100,12 @@ impl pallet_bulletin_transaction_storage::Config for Runtime {
 	type MaxPermanentStorageSize = MaxPermanentStorageSize;
 	type AuthorizationPeriod = AuthorizationPeriod;
 	type AuthorizerRegistrarOrigin = frame_system::EnsureRoot<Self::AccountId>;
-	type Authorizer = EitherOfDiverse<
-		EitherOfDiverse<
+	type Authorizer = EitherOf<
+		EitherOf<
 			// Root can do whatever.
-			crate::EnsureRoot<Self::AccountId>,
+			AsAuthorizer<crate::EnsureRoot<Self::AccountId>, Self::AccountId, crate::BlockNumber>,
```
bkontur, line 106: @franciscoaguirre this `AsAuthorizer` looks really good!

pallets/transaction-storage/src/lib.rs
```diff
@@ -622,22 +626,14 @@ pub mod pallet {
 			transactions: u32,
 			bytes: u64,
 		) -> DispatchResult {
-			let authorization_period_override = Self::authorization_period_override_for(&origin);
-			// Capture the signer (if any) before `ensure_origin` consumes the origin —
-			// we use it post-check to charge their `AllowedAuthorizers` budget. Root /
-			// XCM origins have no signer and no budget to consume.
-			let signer = frame_system::ensure_signed(origin.clone()).ok();
-			T::Authorizer::ensure_origin(origin)?;
+			let maybe_authorizer = T::Authorizer::ensure_origin(origin)?;
```
bkontur, line 629: Yes, yes, this is much better than `let signer = frame_system::ensure_signed`

bkontur (approved):

PR #569, opened 2026-05-28
Per-authorizer feeless flag + System provider management

bkontur:
## Summary

- Adds `feeless: bool` to `AuthorizerBudget` and threads it through `AuthorizationOrigin`, so `#[pallet::feeless_if]` on `authorize_account` and `refresh_account_authorization` can skip the transaction-payment fee when the dispatching signer is an `AllowedAuthorizers` entry with `feeless = true`. Genesis-loaded authorizers default to `feeless: true`; the `v4->v5` migration sets `feeless = true` on existing entries.
- Manages System provider references for `AllowedAuthorizers`: `add_authorizer` (and genesis) `inc_providers`, `remove_authorizer` / `remove_exhausted_authorizer` `dec_providers`. This keeps a `feeless` authorizer's System `Account` alive between dispatches so its replay-protection nonce isn't reset by reaping.


pallets/transaction-storage/src/migrations.rs
```diff
@@ -897,14 +839,14 @@ pub mod v5 {
 			AllowedAuthorizers::<T>::translate::<V4AuthorizerBudget<BlockNumberFor<T>>, _>(
 				|_who, old| {
 					migrated = migrated.saturating_add(1);
-					Some(AuthorizerBudget { quota: old.quota, valid_until: old.valid_until })
+					Some(AuthorizerBudget {
+						quota: old.quota,
+						valid_until: old.valid_until,
+						feeless: true,
```
bkontur, line 850:
```suggestion
						feeless: false,
```

pallets/transaction-storage/src/migrations.rs
```diff
@@ -897,14 +839,14 @@ pub mod v5 {
 			AllowedAuthorizers::<T>::translate::<V4AuthorizerBudget<BlockNumberFor<T>>, _>(
 				|_who, old| {
 					migrated = migrated.saturating_add(1);
-					Some(AuthorizerBudget { quota: old.quota, valid_until: old.valid_until })
+					Some(AuthorizerBudget {
+						quota: old.quota,
+						valid_until: old.valid_until,
+						feeless: false,
```
karolk91:
```suggestion
						feeless: true,
```

pallets/transaction-storage/src/migrations.rs
```diff
@@ -897,14 +839,14 @@ pub mod v5 {
 			AllowedAuthorizers::<T>::translate::<V4AuthorizerBudget<BlockNumberFor<T>>, _>(
 				|_who, old| {
 					migrated = migrated.saturating_add(1);
```
karolk91, line 841: I think we should also inc_providers during migration

pallets/transaction-storage/src/lib.rs
```diff
@@ -2463,6 +2497,16 @@ pub mod pallet {
...
 		}
 
+		/// Backs `#[pallet::feeless_if]` on `authorize_account` and
+		/// `refresh_account_authorization`. Goes through `T::Authorizer::ensure_origin`
+		/// so Root / XCM (`Ok(None)`) are not feeless via this flag.
+		pub(crate) fn is_feeless_authorizer(origin: &OriginFor<T>) -> bool {
+			T::Authorizer::ensure_origin(origin.clone())
+				.ok()
+				.flatten()
+				.is_some_and(|ctx| ctx.feeless)
```
karolk91: should we also check `is_exhausted` here ? to not allow feeless authorizers that don't have budget ? would be also good to add test case for this scenario - regardless of decision made here

pallets/transaction-storage/src/migrations.rs
bkontur: I had this also true, but decided to false, because the authorizer which you added yesterday pays :)

pallets/transaction-storage/src/migrations.rs
bkontur: > I think we should also inc_providers during migration

haha, this is what I removed before I pushed, Paseo authorizer has inc provider because of added PAS,
but ok, I can revert that back

pallets/transaction-storage/src/tests.rs
```diff
@@ -4481,6 +4467,81 @@ fn valid_until_beyond_default_period_does_not_clamp() {
...
+		assert_eq!(providers_of(who), 0);
+
+		// Re-remove must not underflow.
+		assert_ok!(TransactionStorage::remove_authorizer(RuntimeOrigin::root(), who));
+		assert_eq!(providers_of(who), 0);
+
+		// `remove_exhausted_authorizer` path also dec's.
+		AllowedAuthorizers::<Test>::insert(
+			who,
+			AuthorizerBudget {
```
karolk91: just nit: we sometimes use `AuthorizerBudget` directly and sometimes `test_budget` helper - lets maybe be consistent and always use `test_budget`

pallets/transaction-storage/src/migrations.rs
karolk91: I think current authorizer should be feeless so I would say `true` is good here

karolk91:
If this is the last change before v16 release - lets maybe re-run also the benchmarks

pallets/transaction-storage/src/migrations.rs
bkontur: reverted back: https://github.com/paritytech/polkadot-bulletin-chain/pull/569/commits/d88c35739854bf4f593fb3d3f1fab97a068345fd

pallets/transaction-storage/src/lib.rs
bkontur: > should we also check `is_exhausted` here ? to not allow feeless authorizers that don't have budget ? would be also good to add test case for this scenario - regardless of decision made here

good point, I added also is_expired check: https://github.com/paritytech/polkadot-bulletin-chain/pull/569/commits/f455fd0df59d39654d6863c46b17e3c4cac6e535

pallets/transaction-storage/src/tests.rs
bkontur: fixed by: https://github.com/paritytech/polkadot-bulletin-chain/pull/569/commits/53fab828ecbc01b9e08661f107f0a0399a99b50b

karolk91: I see we now use something like
```
AuthorizerBudget { feeless: true, ..test_budget(tx, by) }
```
I was thinking about using just 
`test_budget(tx, by, true)`

but that just some nit

karolk91 (approved):

PR #570, opened 2026-05-28
examples: let chopsticks author past RetentionPeriod

x3c41a:
## Summary
- `check_chopsticks.js` now overrides `TransactionStorage::RetentionPeriod` to `u32::MAX` on the fork and authors several blocks (default 8, override via `CHOPSTICKS_NEW_BLOCKS`).

## Why
Chopsticks can't construct a `TransactionStorageProof` (no off-chain chunks), so authoring past `store_block + RetentionPeriod` panics in `apply_block_inherents` with `Storage proof must be checked once in the block`. With `RetentionPeriod = u32::MAX`, `target_number.saturating_sub(MAX) == 0` for every block, so `proof_ok` short-circuits via the `target_number.is_zero()` branch and no proof is required.

## Verified
- Local zombienet (`retention=10`, store at block 11): control panics at block 21; with override, produced blocks 12–23 cleanly.


examples/check_chopsticks.js
```diff
@@ -1,30 +1,45 @@
...
+
 console.log(`Setting up Chopsticks with Bulletin chain (endpoint: ${endpoint})...`);
 
 try {
-  const chain = await setup({ endpoint });
+  const chain = await setup({ endpoint, mockSignatureHost: true });
   await chain.api.isReady;
   console.log("Chain setup complete.");
 
+  await setStorage(chain, {
```
bkontur, line 17: @x3c41a @karolk91 I just thinking, so basically, imagine somebody wants to use Chopstics to generate couple of blocks like this test is doing, what do they need to do - `setStorage(RetentionPeriod=u32::MAX)`?  Can we do/achieve this with some Chopstics magic (innerProvider, standard Chopstics hack with storage mutation...?

Do we even need this setStorage hack when this is merged and tested: https://github.com/paritytech/smoldot/pull/3263#pullrequestreview-4392247879 ?

karolk91:
We no longer need this change, proof checks are now handled/forged on chopsticks level: https://github.com/AcalaNetwork/chopsticks/pull/1037

bkontur:
> We no longer need this change, proof checks are now handled/forged on chopsticks level: [AcalaNetwork/chopsticks#1037](https://github.com/AcalaNetwork/chopsticks/pull/1037)

closing per :point_up_2:

PR #571, opened 2026-05-28
Split renewal/auto-renewal into a dedicated pallet-data-renewal

### LINKED ISSUE #628: Split renewal/auto-renewal into pallet-data-renewal
_Tracking issue from the cross-repo roadmap plan (2026-06-23)._

Extract renewal/auto-renewal into a dedicated pallet (PR #571). Decide pre- vs post-mainnet given migration cost.

**Covers PR(s):** #571
**Theme:** Renewal & Authorization · **Category:** Refactor
**Priority:** P2 · **Effort:** L (~8 wk) · **Target:** Bulletin Mainnet (Sep '26)

> [!NOTE]
> Relates to #211; large refactor — feature-freeze candidate to defer.

rosarp:
### What

Extracts all data-renewal logic out of `pallet-bulletin-transaction-storage` into a new `pallet-bulletin-data-renewal`:

- **Moved out of the storage pallet:** `force_renew`, `renew`, `enable_auto_renew`, `disable_auto_renew`, `process_pending_renewals`, the `Renewals` (was `AutoRenewals`) / `PendingAutoRenewals` storage, the renewal transaction-extension validation, benchmarks and weights.
- **Cross-pallet contract:** the new pallet sits on top via a direct `Config` bound (no virtual dispatch). The storage pallet exposes a small public API plus an `OnObsoleteTransactions` hook, and keeps two *opaque* payloads the runtime wires — `EntryMeta` (`EntryKind`) and `AuthorizationExtra` (`PermanentExtent`). It has no renewal vocabulary of its own.
- **No direct storage access:** the renewal pallet must not touch this pallet's storage items. `BlockTransactions` and `TransactionByContentHash` stay `pub(super)`; renewal appends through a `BlockTransactionsMut` handle (`Pallet::with_block_transactions`), so the `BlockTransactions` / `TransactionByContentHash` / `MaxBlockTransactions` / host-index invariants stay in the pallet that owns them. Extracted to `main` in #721.
- **Validation:** one `ValidateAuthorizedCalls` extension walks the call tree once and offers each leaf to a tuple of `LeafValidator`s (`StorageLeaves`, `RenewalLeaves`); the first to claim a leaf validates it.
- **Migration:** one-shot relocation of `Renewals`, `PendingAutoRenewals` **and `PermanentStorageUsed`** from the `TransactionStorage` prefix to the new pallet's. `Authorizations` is deliberately untouched — `AuthorizationExtent::extra` occupies the slot the pre-split `bytes_permanent` field had, so existing values already decode as the new layout (pinned by `authorization_encoding_matches_pre_split_layout`).
- **Runtimes:** wired into both `bulletin-westend` and `bulletin-paseo`, with separate weight files.

### Why

`pallet-transaction-storage` had grown to mix two concerns: raw storage/retention, and the renewal/auto-renewal lifecycle. Splitting them keeps each pallet focused, isolates the renewal economics (prepayment, per-cycle charging) behind an explicit contract, and lets renewal be tested, benchmarked and evolved independently.

### Behavioural differences from `main`

The intent is relocation, not new behaviour, but the split is not byte-identical. Nothing here touches state or consensus:

| Area | Change |
|---|---|
| Pool tag prefixes | Prefixes and pricing are now runtime-supplied `ValidTransactionParams` per family. Three of four families keep `main`'s tags; preimage `store` and preimage `force_renew` no longer share one, so both can take a slot in one block against the same preimage authorization. Tags are node-local and recomputed per validation. |
| Rejection codes | `promote` reported a bad blob size as `Custom(0)` (= `IMPOSSIBLE`, "implies a bug"); now `BAD_DATA_SIZE` (`1`). `force_renew` mapped a data-size failure to `TooManyTransactions`; now its own `Error::BadDataSize` (appended last, so existing indices are stable). |
| `try_state` | The `PermanentStorageUsed` accounting equality is enforced under `cfg(test)` and logged under `try-runtime`: live chains drift both ways through history this pallet did not create (uncharged pre-counter `Renew` entries; Root dropping a prepaid registration). |
| `Config` | Pool params split per family across the three pallets. All are `#[pallet::constant]`, so they appear in metadata. |

`integrity_test` asserts the tag prefixes are pairwise distinct — with prefixes now coming from `Config`, a copy-paste could otherwise make two families dedup against each other silently, and `ContentHash` and `AccountId32` both encode to 32 bytes.

### Weights

`on_initialize_with_expiry` needed a `BenchmarkHelper::register_worst_case_entry` hook: the sweep calls `handle_obsolete` once per aged-out entry, which looks up a renewal registration and queues it, and the storage pallet cannot plant that state itself. Without the hook the benchmark measured a sweep with nothing to find. **Weights must be regenerated for this to take effect.**

### Known open

- Westend is on `TransactionStorage` storage version 1, and MBMs run *after* single-block migrations. `v4`/`v5` are wired single-block while `v3` is an MBM, so on Westend `v2` runs (1→2), `v4` and `v5` skip their `== 3` / `== 4` gates, and `v3` then lands it at 3 — `v5`'s `AuthorizerBudget` reshape does not run. Moving `v4` and `v5` into `MbmMigrations` after `v3` fixes it.

---

_Closes #628 — cross-repo roadmap tracking._


rosarp:
/cmd bench --pallet pallet_bulletin_data_renewal pallet_bulletin_transaction_storage

bkontur:
@rosarp please, rebase/merge conflicts

pallets/transaction-storage/src/lib.rs
```diff
@@ -1227,20 +970,13 @@ pub mod pallet {
...
+			// `pallet-bulletin-data-renewal`'s own `ProvideInherent`.
 			let proof = data
 				.get_data::<TransactionStorageProof>(&Self::INHERENT_IDENTIFIER)
 				.unwrap_or(None);
-			let has_pending_renewals = !PendingAutoRenewals::<T>::get().is_empty();
-
-			if proof.is_none() && !has_pending_renewals {
-				return None;
-			}
+			proof.as_ref()?;
```
bkontur: @rosarp what is this doing? Why not join it with:?
```
let proof = data
				.get_data::<TransactionStorageProof>(&Self::INHERENT_IDENTIFIER)
				.unwrap_or(None);
```

pallets/data-renewal/src/lib.rs
```diff
@@ -0,0 +1,562 @@
...
+
+	const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
+
+	#[pallet::pallet]
+	#[pallet::storage_version(STORAGE_VERSION)]
+	pub struct Pallet<T>(_);
+
+	/// Maps content hash to the account that registered it for auto-renewal.
+	#[pallet::storage]
+	pub type AutoRenewals<T: Config> =
```
bkontur: Maybe we could name it now just `Renewals`

```suggestion
	/// Maps content hash to the account that registered it for renewal.
	#[pallet::storage]
	pub type Renewals<T: Config> =
```

pallets/transaction-storage/src/benchmarking.rs
```diff
@@ -403,139 +361,15 @@ mod benchmarks {
...
-	/// so blocks where only one branch has work are conservatively over-charged.
+	/// Worst-case benchmark for the storage-proof inherent (drain has moved to
+	/// `pallet-bulletin-data-renewal`; the linear component `n` is now unused but
+	/// kept for ABI compatibility with the existing `WeightInfo::apply_block_inherents(n)`
+	/// declaration).
 	#[benchmark]
 	fn apply_block_inherents(
 		n: Linear<0, { T::MaxBlockTransactions::get() }>,
 	) -> Result<(), BenchmarkError> {
+		let _ = n;
```
bkontur: hmm, shouldn't we just remove `n: Linear<0, { T::MaxBlockTransactions::get() }>,` ?

```suggestion
```

pallets/data-renewal/src/lib.rs
```diff
@@ -0,0 +1,562 @@
...
+	impl<T: Config> ProvideInherent for Pallet<T> {
+		type Call = Call<T>;
+		type Error = sp_transaction_storage_proof::InherentError;
+		const INHERENT_IDENTIFIER: InherentIdentifier = *b"datarenw";
+
+		fn create_inherent(_data: &InherentData) -> Option<Self::Call> {
+			if PendingAutoRenewals::<T>::get().is_empty() {
+				return None;
+			}
+			Some(Call::process_pending_renewals {})
```
bkontur, line 426: @karolk91 now we have two inherents per block, do you see any complications? Or should we be aware of more validations? Not to blow block size?

Other words, are we still 100% sure, that we don't stop block production?

karolk91: I don't see any issue with 1 vs 2 inherents - just regular concerns about implementation itself and benchmarking done right

pallets/transaction-storage/src/lib.rs
rosarp: Fixed

pallets/transaction-storage/src/lib.rs
```diff
@@ -2607,10 +2137,11 @@ impl<T: Config> Pallet<T> {
...
 
-	/// Verify the chain-wide permanent-storage accounting invariants:
-	/// - `PermanentStorageUsed == Σ Renew sizes in Transactions + Σ paid AutoRenewals sizes` — the
-	///   paid term covers the prepayment window between `renew` / `enable_auto_renew` charging the
-	///   counter and `do_process_auto_renewals` writing the `Renew` entry.
+	/// Verify the chain-wide permanent-storage accounting invariants visible from
+	/// this pallet alone:
+	/// - `PermanentStorageUsed >= Σ Renew sizes in Transactions` — the chain-wide counter is at
+	///   least the on-chain renewed bytes. The remainder is accounted for by paid auto-renewal
+	///   registrations and checked in `pallet-bulletin-data-renewal`'s own `try_state`.
```
karolk91: > registrations and checked in `pallet-bulletin-data-renewal`'s own `try_state`.

there doesn't seem to be any `try_state` in new pallet

pallets/transaction-storage/src/lib.rs
```diff
@@ -2728,20 +2247,12 @@ pub fn ensure_weight_sanity<T: Config>(collator_pov_percent: Option<u64>) {
...
-	// 4. MaxBlockTransactions renew calls must fit by ref_time.
-	let renew_weight = T::WeightInfo::renew().saturating_add(base_extrinsic);
-	let total_renew_ref_time = renew_weight.ref_time().saturating_mul(max_block_txs as u64);
-	assert!(
-		total_renew_ref_time <= effective_normal.ref_time(),
-		"MaxBlockTransactions ({max_block_txs}) renew calls: \
-		 total ref_time {total_renew_ref_time} exceeds effective normal limit {}",
-		effective_normal.ref_time(),
-	);
+	// 4. Renew-call ref-time fit is checked by `pallet-bulletin-data-renewal`'s own
```
karolk91: I don't see `ensure_weight_sanity` in new pallet

pallets/data-renewal/Cargo.toml
```diff
@@ -0,0 +1,52 @@
+[package]
+name = "pallet-bulletin-data-renewal"
```
karolk91, line 2: I would suggest maybe the name to include `transaction-storage` as its not some generic "data" renewal logic and rather coupled with transaction storage
```suggestion
name = "pallet-bulletin-transaction-storage-renewal"
```

pallets/transaction-storage/src/lib.rs
```diff
@@ -2542,9 +2072,9 @@ pub mod pallet {
 
 pub mod extension;
 
-#[cfg(any(test, feature = "try-runtime"))]
+#[allow(dead_code)] // exercised by `try_state` (try-runtime) and cross-pallet tests
```
karolk91:
```suggestion
#[cfg(any(test, feature = "std", feature = "try-runtime"))]
```

pallets/data-renewal/src/migrations.rs
```diff
@@ -0,0 +1,171 @@
...
+
+		// `AutoRenewals` (StorageMap): re-key every entry from the old
+		// `TransactionStorage` prefix to the new `DataRenewal` prefix, reshaping any
+		// pre-v4 `{ account }` value into the current `RenewalData { account,
+		// recurring, paid }` layout. A plain `move_prefix` would carry pre-v4 bytes
+		// over verbatim and leave them undecodable under this pallet's type — see the
+		// retired `transaction_storage::migrations::v4::MigrateV3ToV4` for the original
+		// reshape. The Blake2_128Concat key suffix (the content hash) is identical
+		// across prefixes, so only the prefix is rewritten.
+		let old_auto_prefix = storage_prefix(OLD_PALLET, b"AutoRenewals");
```
karolk91:
```suggestion
		let old_pallet = <txs::Pallet<T> as PalletInfoAccess>::name().as_bytes();
		let old_auto_prefix = storage_prefix(old_pallet, b"AutoRenewals");
```

pallets/data-renewal/src/migrations.rs
```diff
@@ -0,0 +1,171 @@
...
+//! One-shot relocation: moves `AutoRenewals` and `PendingAutoRenewals` from
+//! the legacy `TransactionStorage::*` storage prefix to `DataRenewal::*`.
+//! `PermanentStorageUsed` continues to live in the storage pallet and needs no
+//! relocation.
+
+extern crate alloc;
+
+use crate::{Config, RenewalData};
+use codec::{Decode, Encode};
+use polkadot_sdk_frame::deps::{
```
karolk91, line 32:
```suggestion
use pallet_bulletin_transaction_storage as txs;
use polkadot_sdk_frame::deps::{
```

pallets/data-renewal/src/migrations.rs
```diff
@@ -0,0 +1,171 @@
...
+		// `AutoRenewals` (StorageMap): re-key every entry from the old
+		// `TransactionStorage` prefix to the new `DataRenewal` prefix, reshaping any
+		// pre-v4 `{ account }` value into the current `RenewalData { account,
+		// recurring, paid }` layout. A plain `move_prefix` would carry pre-v4 bytes
+		// over verbatim and leave them undecodable under this pallet's type — see the
+		// retired `transaction_storage::migrations::v4::MigrateV3ToV4` for the original
+		// reshape. The Blake2_128Concat key suffix (the content hash) is identical
+		// across prefixes, so only the prefix is rewritten.
+		let old_auto_prefix = storage_prefix(OLD_PALLET, b"AutoRenewals");
+		let new_auto_prefix = storage_prefix(NEW_PALLET, b"Renewals");
```
karolk91:
```suggestion
		let new_auto_prefix = crate::Renewals::<T>::final_prefix();
```

karolk91:
sorry @rosarp, will disable auto-merge for now - I want to do more testing and will re-enable when done

pallets/data-renewal/src/migrations.rs
```diff
@@ -0,0 +1,173 @@
...
+			new_key.extend_from_slice(&key[old_auto_prefix.len()..]);
+			sp_io::storage::set(&new_key, &value);
+			sp_io::storage::clear(&key);
+			moved = moved.saturating_add(1);
+		}
+
+		// `PendingAutoRenewals` (StorageValue): transient per-block scratch, normally
+		// empty across an upgrade. Move verbatim if present.
+		let old_pending_key = storage_prefix(OLD_PALLET, b"PendingAutoRenewals");
+		let new_pending_key = storage_prefix(NEW_PALLET, b"PendingAutoRenewals");
```
x3c41a: `NEW_PALLET` has to match the name the runtime registers this pallet under; if it ever differs, `PendingAutoRenewals` gets relocated into a dead prefix. Derive the key from the storage item itself, like the `Renewals` move above does with `final_prefix()`:

```suggestion
		let new_pending_key = crate::PendingAutoRenewals::<T>::hashed_key();
```

Alternatively the whole pending move could be dropped: migrations run before `on_initialize` and the old runtime's `on_finalize` asserts the vec is drained every block, so it is always empty here.

pallets/data-renewal/src/migrations.rs
```diff
@@ -0,0 +1,173 @@
...
+	#[cfg(feature = "try-runtime")]
+	fn post_upgrade(
+		state: alloc::vec::Vec<u8>,
+	) -> Result<(), polkadot_sdk_frame::deps::sp_runtime::TryRuntimeError> {
+		use polkadot_sdk_frame::prelude::ensure;
+		let pre = u64::decode(&mut &state[..]).map_err(|_| "pre_upgrade state decode failed")?;
+
+		// Every relocated entry must live under the new prefix and decode as the
+		// current `RenewalData` layout (catches a pre-v4 entry that wasn't reshaped).
+		let new_auto_prefix = storage_prefix(NEW_PALLET, b"Renewals");
```
x3c41a: Same hazard, and here it can bite: `on_runtime_upgrade` writes under `Renewals::final_prefix()` but this check reads the hardcoded name, so the two can diverge.

```suggestion
		let new_auto_prefix = crate::Renewals::<T>::final_prefix();
```

pallets/data-renewal/src/migrations.rs
```diff
@@ -0,0 +1,173 @@
...
+		traits::{Get, GetStorageVersion, OnRuntimeUpgrade, PalletInfoAccess, StorageVersion},
+		weights::Weight,
+	},
+	sp_io,
+};
+
+const LOG_TARGET: &str = "runtime::data-renewal::migrations";
+
+const OLD_PALLET: &[u8] = b"TransactionStorage";
+const NEW_PALLET: &[u8] = b"DataRenewal";
```
x3c41a: Unused once both new-prefix keys are derived from the storage items.

```suggestion
```

x3c41a (commented):
Hardcoded pallet names in the relocation migration. The three suggestions below go together (batch-commit).

karolk91:
@rosarp, please add a test case based on chopsticks that ensures the new inherent works with chopsticks - my understanding is that it should fail as of today because we need to add new inherent support as was added in: https://github.com/AcalaNetwork/chopsticks/pull/1037

and in general please add chopsticks coverage for all proof, auto renew related inherents

pallets/transaction-storage/src/lib.rs
franciscoaguirre, line 210: What about this one?

franciscoaguirre: Should we keep this in this pallet or move it to the new one?

pallets/transaction-storage/src/lib.rs
```diff
@@ -1457,7 +1047,7 @@ pub mod pallet {
 		/// Centralising read + write + events in one helper guarantees every change to the
 		/// chain-wide counter is observable off-chain, and that the near-cap signal fires
 		/// exactly once per crossing.
-		fn update_permanent_storage_used(f: impl FnOnce(u64) -> u64) {
+		pub fn update_permanent_storage_used(f: impl FnOnce(u64) -> u64) {
 			let old = PermanentStorageUsed::<T>::get();
```
franciscoaguirre: This should be moved to the new pallet, no? It only makes sense with renewals

pallets/data-renewal/src/extension.rs
```diff
@@ -0,0 +1,381 @@
...
+		})
+	}
+}
+
+impl<T: Config, I> ValidateBulletinCalls<T, I> {
+	/// Storage-pallet calls that must arrive as direct extrinsics (rejected at
+	/// any wrapper depth > 0). Management calls (`authorize_*`, `refresh_*`,
+	/// `remove_*`) are intentionally allowed inside wrappers.
+	fn is_direct_only_storage_call(call: &txs::Call<T>) -> bool {
+		matches!(call, txs::Call::store { .. } | txs::Call::store_with_cid_config { .. })
```
franciscoaguirre: Shouldn't this be in the storage pallet?

pallets/data-renewal/src/lib.rs
```diff
@@ -0,0 +1,562 @@
...
+//!   (per-block scratch queue, drained by the inherent).
+//!
+//! ## Cross-pallet contract
+//!
+//! - **Down → storage:** dispatchables and the trait callback read/mutate `Transactions`,
+//!   `TransactionByContentHash`, `BlockTransactions`, and `Authorizations` directly through
+//!   `pallet_bulletin_transaction_storage`'s public API.
+//! - **Up ← storage:** [`OnObsoleteTransactions::handle_obsolete`] is called by the storage
+//!   pallet's `on_initialize` when entries age out at the `RetentionPeriod` boundary; entries with
+//!   an `Renewals` registration are pushed to [`PendingAutoRenewals`] for the same block's inherent
```
franciscoaguirre:
```suggestion
//!   a `Renewals` registration are pushed to [`PendingAutoRenewals`] for the same block's inherent
```

pallets/data-renewal/src/lib.rs
```diff
@@ -0,0 +1,562 @@
...
+//!   to drain.
+//! - **Per-cycle accounting** (per-account `bytes_permanent` and the chain-wide
+//!   `PermanentStorageUsed`) is charged by the storage pallet's `check_authorization` with
+//!   `is_renew = true`.
```
franciscoaguirre: This should be in this pallet

pallets/data-renewal/src/lib.rs
```diff
@@ -0,0 +1,562 @@
...
+					});
+				} else {
+					if charged {
+						// Reverse the chain-wide `PermanentStorageUsed` bump that
+						// `check_authorization` applied for this cycle. The per-account
+						// `bytes_permanent` / `transactions` increments are intentionally
+						// left burned: slot-cap rejection at inherent time is a chain-level
+						// pathological event.
+						let size_u64: u64 = tx_info.size.into();
+						pallet_bulletin_transaction_storage::Pallet::<T>::update_permanent_storage_used(
```
franciscoaguirre: That's why it would be better to have that storage in this pallet

pallets/data-renewal/src/migrations.rs
```diff
@@ -0,0 +1,173 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! One-shot relocation: moves `AutoRenewals` and `PendingAutoRenewals` from
+//! the legacy `TransactionStorage::*` storage prefix to `DataRenewal::*`.
+//! `PermanentStorageUsed` continues to live in the storage pallet and needs no
+//! relocation.
```
franciscoaguirre: Why though?

franciscoaguirre (approved):
I'm fine with the changes. I just feel like it splits some stuff but not everything. pallet-transaction-storage still knows the concept of renewals

runtimes/bulletin-westend/src/lib.rs
```diff
@@ -154,10 +156,8 @@ pub mod migrations {
 	pub type SingleBlockMigrations = (Unreleased, Permanent);
 
 	/// MBM migrations to apply on runtime upgrade.
-	pub type MbmMigrations = (
-		pallet_bulletin_transaction_storage::migrations::v3::MigrateV2ToV3<Runtime>,
-		pallet_bulletin_transaction_storage::migrations::v4::MigrateV3ToV4<Runtime>,
-	);
+	pub type MbmMigrations =
+		(pallet_bulletin_transaction_storage::migrations::v3::MigrateV2ToV3<Runtime>,);
```
franciscoaguirre, line 168: Just checked, MBMs always run **after** single-block migrations. This means we will have storage versions issues here. Could we move everything to Mbm even if it won't take up more than a block?

franciscoaguirre (changes requested):

franciscoaguirre (approved):
I checked the storage version on each chain and it already run all the migrations. Still, it would be better to reorder them

bkontur:
@x3c41a @antkve @RafalMirowski1  guys, please review, (no Claude-submitted comments please)


@rosarp please, resolve conflicts and rebase

rosarp:
/cmd bench --pallet pallet_bulletin_data_renewal pallet_bulletin_transaction_storage

rosarp:
/cmd bench --pallet pallet_bulletin_transaction_storage_renewal pallet_bulletin_transaction_storage

pallets/data-renewal/src/lib.rs
rosarp: outdated

pallets/data-renewal/Cargo.toml
```diff
@@ -0,0 +1,54 @@
+[package]
+name = "pallet-bulletin-transaction-storage-renewal"
```
bkontur: I don't know, this is kind of inconsistent naming — why not just `pallet-bulletin-data-renewal`?

For `pallet-bulletin-transaction-storage-renewal` I would rather expect the directory to be `pallets/transaction-storage-renewal`, but it is `pallets/data-renewal`.

```suggestion
name = "pallet-bulletin-data-renewal"
```

Then we could also avoid doing:

```rust
use pallet_bulletin_transaction_storage_renewal as txs_renewal;
```

`txs_renewal` brings some confusion (for me at least) — are we doing "transactions renewal"? Which is kind of not 100% correct, because we are renewing content as well; yeah, content which was generated by a transaction, but still.

I think all the time we are using the "data renewal" words, right? The runtime registers it as `DataRenewal`, the storage prefix is `DataRenewal`, the log targets are `runtime::data-renewal` / `runtime::data-renewal::migrations`, the directory is `pallets/data-renewal`, and the pallet docs open with "Data-renewal layer for the Bulletin chain". The crate name is the only place that says "transaction storage renewal", and it is also the reason we need the `txs_renewal` alias in 6 files.

What do you think?

rosarp: Earlier it was `pallet-bulletin-data-renewal`.
In [comment](https://github.com/paritytech/polkadot-bulletin-chain/pull/571#discussion_r3569294619) at that point it made sense. 
I agree its a inconsistent naming now with folder structure , pallet naming & many other places which I missed!
I think either `data` or `store` is short and sweet. 
As we are already using DataRenewal many places, i could revert it back to `pallet-bulletin-data-renewal`.
Otherwise there will be many places to be updated for naming consistency.

pallets/hop-promotion/Cargo.toml
```diff
@@ -28,6 +28,7 @@ bulletin-transaction-storage-primitives = { workspace = true }
 
 [dev-dependencies]
 bulletin-pallets-common = { workspace = true, features = ["std"] }
+pallet-bulletin-transaction-storage-renewal = { workspace = true, features = ["std"] }
```
bkontur: I don't think that `hop-promotion` should anyhow depend on renewals pallet.
I see this is just because of `promote_has_lower_priority_than_store_and_renew` check, but this check should be done on the **runtime-level** the integrity_test/try-state or unit-test in the runtime directly.

I am trying to refactor this as a part of: https://github.com/rosarp/polkadot-bulletin-chain/pull/12

pallets/data-renewal/src/lib.rs
```diff
@@ -0,0 +1,858 @@
...
+		if pallet_bulletin_transaction_storage::Pallet::<T>::block_transactions_full() {
+			return Err(InvalidTransaction::ExhaustsResources.into());
+		}
+		Self::check_renew_authorization(
+			&AuthorizationScope::Preimage(info.content_hash),
+			info.size,
+			context.consume_authorization(),
+		)?;
+		Ok(context.want_valid_transaction().then(|| {
+			pallet_bulletin_transaction_storage::Pallet::<T>::preimage_store_renew_valid_transaction(
```
bkontur: probably this is part of the @franciscoaguirre's comment, that still tx pallet knows renew concept, I agree, if we are doing this split, we should do more stuff:

I am trying to refactor this as a part of: https://github.com/rosarp/polkadot-bulletin-chain/pull/12

pallets/transaction-storage/src/types.rs
```diff
@@ -116,39 +115,76 @@ pub type AuthorizedCallerFor<T> = AuthorizedCaller<<T as frame_system::Config>::
...
-pub(crate) type AuthorizationFor<T> = Authorization<BlockNumberFor<T>>;
+pub type AuthorizationFor<T> = Authorization<BlockNumberFor<T>, <T as Config>::AuthorizationExtra>;
 
-/// Distinguishes a stored transaction created by `store` (temporary) from one created by
-/// `renew` (permanent), so that `on_initialize`'s obsolete-block cleanup can decrement
-/// `PermanentStorageUsed` only for the renewed entries.
-#[derive(
-	Copy, Clone, PartialEq, Eq, Debug, Encode, Decode, scale_info::TypeInfo, MaxEncodedLen,
-)]
-pub enum TransactionKind {
```
bkontur, line 142: @rosarp @karolk91 I don't know, what do we gain with: `pub enum EntryKind {`, it is basically the same enum, neither is perferct "tx pallet: Store+Renew" now "renewal pallet Store+Renew". I would maybe keep `TransactionKind` in the primitives and do not change this (but I keep reviewing)

rosarp: It was added by @franciscoaguirre  in Split pallet improvements ([#11](https://github.com/rosarp/polkadot-bulletin-chain/pull/11))

bkontur:
/cmd bench --pallet pallet_bulletin_data_renewal pallet_bulletin_transaction_storage

runtimes/bulletin-paseo/src/storage.rs
```diff
@@ -98,13 +124,23 @@ impl pallet_bulletin_transaction_storage::Config for Runtime {
...
-	type StoreRenewLongevity = StoreRenewLongevity;
-	type RemoveExpiredAuthorizationPriority = RemoveExpiredAuthorizationPriority;
-	type RemoveExpiredAuthorizationLongevity = RemoveExpiredAuthorizationLongevity;
+	type StoreTxParams = StoreTxParams;
+	type AuthorizeTxParams = AuthorizeTxParams;
+	type RemoveExpiredAccountAuthorizationTxParams = RemoveExpiredAccountAuthorizationTxParams;
+	type RemoveExpiredPreimageAuthorizationTxParams = RemoveExpiredPreimageAuthorizationTxParams;
+	type RemoveExhaustedAuthorizerTxParams = RemoveExhaustedAuthorizerTxParams;
+	type EntryMeta = pallet_bulletin_data_renewal::EntryKind;
+	type AuthorizationExtra = pallet_bulletin_data_renewal::PermanentExtent;
```
bkontur, line 125: @franciscoaguirre it looks like EntryKind/PermanentExtent are clean enum/structs not tight to the pallet's Config, so it should work without `impl pallet_bulletin_data_renewal::Config for Runtime {` (e.g. for prod)

franciscoaguirre: True, we just need the types in the config of pallet-transaction-storage

bkontur:
> @rosarp, please add a test case based on chopsticks that ensures the new inherent works with chopsticks - my understanding is that it should fail as of today because we need to add new inherent support as was added in: [AcalaNetwork/chopsticks#1037](https://github.com/AcalaNetwork/chopsticks/pull/1037)
> 
> and in general please add chopsticks coverage for all proof, auto renew related inherents

@karolk91 @rosarp it was added, but not as a part of CI, when I added it there it fails (didn't investigate yet):
https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/30497103441/job/90728678180?pr=571
https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/30497103441/job/90728678180?pr=571

I know that we can fix it probably here: https://github.com/AcalaNetwork/chopsticks/pull/1037
but this brings me back to the "1 or 2 inherents problem", basically, this PR is combination of "tighly-coupling"  and "loose-coupling" (PolkadotSDK uses both patterns, so no problem with either).

What do we gain from two inherents? Does it just complicate block processing? Can we just have 1 inherent and inject renewals draining inherent functionality over some trait like `OnObsoleteTransactions`? (At least, it would be easier for external tooling - need to modify chopstics?...), I mean, looks like 2 inherents should also be good and working, but let's make well-informed decisions here - @karolk91 @rosarp                                                                                                                                                              

--- 
So:
- [ ] 1 inherent with callback
- [ ] 2 inherents + fix Chopstics also

bkontur:
/cmd bench --pallet pallet_bulletin_data_renewal pallet_bulletin_transaction_storage

franciscoaguirre:
@bkontur I'd go with two inherents, it fits the split and we gain simplicity. If we had one inherent with a callback we need to add more code to that and make sure that works properly. The two inherents work independently. It feels like the right approach. We have two pallets, we have two inherents that do independent things

franciscoaguirre:
I created the chopsticks fix here: https://github.com/AcalaNetwork/chopsticks/pull/1064

karolk91:
> > @rosarp, please add a test case based on chopsticks that ensures the new inherent works with chopsticks - my understanding is that it should fail as of today because we need to add new inherent support as was added in: [AcalaNetwork/chopsticks#1037](https://github.com/AcalaNetwork/chopsticks/pull/1037)
> > and in general please add chopsticks coverage for all proof, auto renew related inherents
> 
> @karolk91 @rosarp it was added, but not as a part of CI, when I added it there it fails (didn't investigate yet): https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/30497103441/job/90728678180?pr=571 https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/30497103441/job/90728678180?pr=571
> 
> I know that we can fix it probably here: [AcalaNetwork/chopsticks#1037](https://github.com/AcalaNetwork/chopsticks/pull/1037) but this brings me back to the "1 or 2 inherents problem", basically, this PR is combination of "tighly-coupling" and "loose-coupling" (PolkadotSDK uses both patterns, so no problem with either).
> 
> What do we gain from two inherents? Does it just complicate block processing? Can we just have 1 inherent and inject renewals draining inherent functionality over some trait like `OnObsoleteTransactions`? (At least, it would be easier for external tooling - need to modify chopstics?...), I mean, looks like 2 inherents should also be good and working, but let's make well-informed decisions here - @karolk91 @rosarp
> 
> So:
> 
>     * [ ]  1 inherent with callback
> 
>     * [ ]  2 inherents + fix Chopstics also

We wanted this change to separate renew concepts as much as possible so I think 2 inherents is the right way

pallets/transaction-storage/src/benchmarking.rs
```diff
@@ -36,6 +36,20 @@ pub trait BenchmarkHelper<T: Config> {
...
 	/// built with `random_hash` as randomness.
 	fn encoded_check_proof(random_hash: &[u8]) -> Vec<u8>;
+
+	/// `EntryMeta` value that makes the `on_initialize` expiry sweep's
+	/// `OnObsoleteTransactions` callback do its worst-case work. Runtimes wiring the
+	/// renewal pallet should return its permanent variant so the benchmark exercises
+	/// the renewed-byte counter decrement.
+	fn worst_case_entry_meta() -> T::EntryMeta {
+		Default::default()
+	}
```
karolk91, line 46: I would remove the default implementation in the trait and move it to `DefaultCheckProofHelper` instead so it can't be silently missed by implementers
```suggestion
	fn worst_case_entry_meta() -> T::EntryMeta;
```

pallets/transaction-storage/src/benchmarking.rs
```diff
@@ -36,6 +36,20 @@ pub trait BenchmarkHelper<T: Config> {
...
+
+	/// Give `content_hash` the consumer-pallet state that makes
+	/// [`Config::OnObsoleteTransactions`] take its most expensive path. Without it the expiry
+	/// sweep is benchmarked with no registration to find, so the lookup and the queueing it
+	/// leads to go uncharged.
+	fn register_worst_case_entry(_content_hash: ContentHash) {}
```
karolk91, line 52:
```suggestion
	fn register_worst_case_entry(_content_hash: ContentHash);
```

examples/check_auto_renew_chopsticks.js
```diff
@@ -2,16 +2,28 @@
...
- *  6. Jumps to the expiry block and triggers on_initialize
- *  7. Verifies PendingAutoRenewals was populated (proven by on_finalize panic)
+ *  2. Sets up authorization + a stored transaction via PAPI-encoded storage
+ *  3. Tests enable_auto_renew: sets DataRenewal.Renewals entry, verifies it
+ *  4. Tests disable_auto_renew: removes the entry, verifies it's cleared
+ *  5. Re-enables auto-renewal for the expiry test
+ *  6. Advances to the expiry block. There `TransactionStorage::on_initialize`
+ *     ages the data out and queues it into `DataRenewal.PendingAutoRenewals`
+ *     for the same block's `process_pending_renewals` mandatory inherent.
+ *  7. Injects the `DataRenewal.process_pending_renewals` inherent into that
```
karolk91, line 17: The test should not inject `DataRenewal.process_pending_renewals` on itself - we should expect chopsticks to do this, but we can keep like this for now until https://github.com/AcalaNetwork/chopsticks/pull/1064 is released. So we can remove this inject part while bumping chopsticks (cc @franciscoaguirre )

karolk91 (approved):

examples/check_auto_renew_chopsticks.js
franciscoaguirre: 100% agree, could be a follow-up

mudigal: Should we atleast open an issue to track the follow-up?

karolk91: addressed in the PR: https://github.com/paritytech/polkadot-bulletin-chain/pull/737

PR #574, opened 2026-05-29
Fix transaction_index issue with HOP promote

franciscoaguirre:
`sp_io::transaction_index::index()` requires the last parameter of the extrinsic to be the data. This is the case with `TransactionStorage::store()` but was not the case with `HopPromotion::promote()`. I reordered the parameters and added a test to make sure we catch something like this in the future


bkontur (approved):

bkontur:
I am adding Also zombienetsdk fór this

bkontur:
I changed the base of this PR to this PR: https://github.com/paritytech/polkadot-bulletin-chain/pull/577.
577 should have failing hop zombienet test
and this PR should pass this test,
this will be ultimate double check.

karolk91 (approved):

PR #575, opened 2026-05-29
HOP Submit signature v2

### LINKED ISSUE #626: HOP submit signature v2
_Tracking issue from the cross-repo roadmap plan (2026-06-23)._

Track HOP submit signature v2 (PR #575). Mainnet-blocking if HOP ships with Bulletin mainnet.

**Covers PR(s):** #575
**Theme:** HOP · **Category:** Feature
**Priority:** P1 · **Effort:** M (~4 wk) · **Target:** Bulletin Mainnet (Sep '26)

> [!NOTE]
> Could instead be filed under existing #544 (HOP follow-ups) — pick one.

franciscoaguirre:
Submit signature V2 adds the recipient list and the genesis hash to the signature

---

_Closes #626 — cross-repo roadmap tracking._


pallets/hop-promotion/src/lib.rs
```diff
@@ -202,5 +309,35 @@ pub mod pallet {
...
+		#[pallet::weight_of_authorize(
+			<T as Config>::WeightInfo::authorize_promote_v2(data.len() as u32, recipients.len() as u32)
+		)]
+		pub fn promote_v2(
+			origin: OriginFor<T>,
+			data: Vec<u8>,
+			_signer: MultiSigner,
+			_signature: MultiSignature,
+			_submit_timestamp: u64,
+			recipients: BoundedVec<MultiSigner, RecipientsBound>,
```
bkontur: `data` at the end :upside_down_face: 

```suggestion
			_signer: MultiSigner,
			_signature: MultiSignature,
			_submit_timestamp: u64,
			recipients: BoundedVec<MultiSigner, RecipientsBound>,
			data: Vec<u8>,
```

@karolk91 thinking if we could enforce this by some macro magic or something :D, that when `tx:index` used inside extrinsic, the data should be the last :D

franciscoaguirre: Yes I made this code before the fix 😂 That's why I marked it as draft. I also thought about using a macro. Will give it a try

pallets/hop-promotion/src/lib.rs
```diff
@@ -176,6 +222,67 @@ pub mod pallet {
...
+		/// signature against the V2 payload, which additionally binds the chain's
+		/// genesis hash (cross-chain replay) and the recipients list (rebinding
+		/// to a different audience).
+		#[allow(clippy::ptr_arg)]
+		pub fn authorize_promote_v2(
+			source: TransactionSource,
+			signer: &MultiSigner,
+			signature: &MultiSignature,
+			submit_timestamp: &u64,
+			recipients: &BoundedVec<MultiSigner, RecipientsBound>,
```
x3c41a: One small mismatch: sc-hop rejects submissions with zero recipients (NoRecipients), but promote_v2 accepts an empty list, and authorize_v2_accepts_empty_recipients even tests for that. 

Not a security problem, since the signature still has to cover the empty list. But if this is not on purpose, it would be good to require at least one recipient here too, so both sides agree.

x3c41a (approved):

dr333ws:
Another thought: the runtime never reads the recipients list itself, it only uses its hash (`blake2_256(recipients.encode())`). So `promote_v2` could take a 32-byte `recipients_hash` instead of the full list. That would save up to ~8.4 KB of block space per promotion (256 recipients x 33 bytes) and drop the per-recipient weight term. Keeping the full list only makes sense if you want the audience to be visible on chain. If that is the goal, a short comment saying so would help, so nobody strips it out later.

pallets/hop-promotion/src/lib.rs
```diff
@@ -59,9 +70,38 @@ pub fn signing_payload(data_hash: &[u8; 32], submit_timestamp: u64) -> [u8; 32]
...
 
+/// Reconstructs the V2 signing payload. Extends V1 with the chain's genesis
+/// hash and a hash of the SCALE-encoded recipients list.
+///
+/// Byte layout (pre-hash):
+/// `HOP_SUBMIT_CONTEXT_V2 || data_hash || submit_timestamp.to_le_bytes()
+///  || genesis_hash || recipients_hash`
+///
+/// Must remain byte-identical to the SDK-side construction in `sc-hop`.
+pub fn signing_payload_v2(
```
x3c41a, line 95: The sc-hop side of this does not exist yet. I could not find `hop-submit-v2:` anywhere in the org, and no matching polkadot-sdk PR. Merging chain-first is fine, but please link the SDK PR in the description once it exists.

One thing to watch out for in that SDK work: sc-hop's `RecipientVec` is a `BoundedVec<Recipient>`, and `Recipient` also carries ack state. The pallet here hashes a plain `BoundedVec<MultiSigner>`. So the SDK has to encode just the `MultiSigner` list, in submission order, not its own type, or the hashes will never match.

A shared test vector (fixed keys, fixed payload, expected hash) in both repos would pin the byte layout down.

pallets/hop-promotion/src/lib.rs
```diff
@@ -176,6 +222,67 @@ pub mod pallet {
...
+			signature: &MultiSignature,
+			submit_timestamp: &u64,
+			recipients: &BoundedVec<MultiSigner, RecipientsBound>,
+			data: &Vec<u8>,
+		) -> Result<(ValidTransaction, Weight), TransactionValidityError> {
+			if matches!(source, TransactionSource::External) {
+				return Err(InvalidTransaction::Call.into());
+			}
+			if !pallet_bulletin_transaction_storage::Pallet::<T>::data_size_ok(data.len()) {
+				return Err(InvalidTransaction::Custom(0).into());
```
bkontur: why `custom(0)`? we can use BAD_DATA_SIZE from tx pallet for this

bkontur:
@franciscoaguirre do we have a counterpart PR for node HOP to use this?

bkontur:
> @franciscoaguirre do we have a counterpart PR for node HOP to use this?

also misses `create_promotion_extrinsic` adjustment, right?

franciscoaguirre:
> @franciscoaguirre do we have a counterpart PR for node HOP to use this?

We can create that later, once the runtime is upgraded to include the v2 signature

bkontur:
> Command "bench --pallet pallet-bulletin-hop-promotion" has failed ❌! [See logs here](https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/30549146449/job/90892955980)

Looks like this failed for bulletin-polkadot, pushed fix: https://github.com/paritytech/polkadot-bulletin-chain/commit/e551d7bb321d00af3e3c6784aed52bbdf7a75868 and reruning this as a part of https://github.com/paritytech/polkadot-bulletin-chain/pull/719

bkontur:
> > Command "bench --pallet pallet-bulletin-hop-promotion" has failed ❌! [See logs here](https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/30549146449/job/90892955980)
> 
> Looks like this failed for bulletin-polkadot, pushed fix: [e551d7b](https://github.com/paritytech/polkadot-bulletin-chain/commit/e551d7bb321d00af3e3c6784aed52bbdf7a75868) and reruning this as a part of #719

actually, the problem is wrong `--pallet pallet-bulletin-hop-promotion` vs `--pallet pallet_bulletin_hop_promotion`

Issue #595, opened 2026-06-10
Codebase audit: licensing, attributions, README, and publish-readiness


mudigal:
## Overview

Comprehensive audit of the Bulletin Chain codebase to ensure publish-readiness — covering licensing consistency, file-level attributions, README completeness, and repository hygiene. This issue serves as the umbrella tracker with existing issues linked as sub-tasks.

Based on [Parity's Publish Checklist](https://docs.google.com/document/d/1gq3kGN5t7DHX5KFZAlJL5W6e8jfRQWdLzbi3sKB4en0/edit?tab=t.0) and prior work in #500.

---

## 1. License & Attribution Alignment

PR #500 established the licensing policy:
- **GPL-3.0-only** for runtimes, console-UI
- **Apache-2.0** for pallets, SDKs, libraries

**However, file-level SPDX headers were never updated to match.** Current state:

| Package | `Cargo.toml` / `package.json` | File SPDX Header | Status |
|---|---|---|---|
| `runtimes/bulletin-westend/` | `GPL-3.0-only` | `Apache-2.0` | **MISMATCH — 30+ files** |
| `runtimes/bulletin-paseo/` | `GPL-3.0-only` | `Apache-2.0` | **MISMATCH — 30+ files** |
| `pallets/*` | `Apache-2.0` | `Apache-2.0` | OK |
| `sdk/rust/` | `Apache-2.0` | `GPL-3.0-or-later WITH Classpath-exception-2.0` | **MISMATCH** |
| `sdk/typescript/` | `Apache-2.0` | `GPL-3.0-or-later WITH Classpath-exception-2.0` | **MISMATCH** |
| `console-ui/` | *(none in package.json)* | *(no headers)* | **MISSING entirely** |

### Action items
- [x] Update all runtime `.rs` file headers from `Apache-2.0` to `GPL-3.0-only`
- [x] Update SDK Rust file headers from `GPL-3.0-or-later WITH Classpath-exception-2.0` to `Apache-2.0`
- [x] Update SDK TypeScript file headers from `GPL-3.0-or-later WITH Classpath-exception-2.0` to `Apache-2.0`
- [x] Add license field and file headers to `console-ui/`
- [x] Verify root `LICENSE` file (currently GPL-3.0) is appropriate, or add per-crate LICENSE files
- [x] Ensure `"This file is part of Substrate"` references are updated where applicable (e.g., `pallets/transaction-storage/src/lib.rs`)

---

[...]

PR #600, opened 2026-06-11
Update README with audit and usage warnings

mudigal:
#595 

Added warning about the audit status and development stage.


waytwotall (approved):
I think this is important to have

PR #611, opened 2026-06-17
Drop historical migrations from paseo runtime

bkontur:
## Summary

Paseo is a fresh chain, so the historical migrations carried over from the westend runtime do not apply to it. This empties both migration tuples in the paseo runtime:

- `Unreleased` (single-block version-bump migrations) → `()`
- `MbmMigrations` (transaction-storage `v2->v3`, `v3->v4`) → `()`

`pallet_migrations` stays wired in (construct_runtime index 7, `frame_system::MultiBlockMigrator`) for future use — no pallet-index/metadata churn.

`Permanent` migrations (`MigrateToLatestXcmVersion`, `SetRetentionPeriodIfZero`) are left intact, as those are idempotent checks meant to run on every upgrade rather than historical version bumps.

## Test

- `cargo check -p bulletin-paseo-runtime`
- `cargo clippy -p bulletin-paseo-runtime --all-targets --all-features -- -D warnings`
- `cargo +nightly fmt --all`

### DIFF
--- a/runtimes/bulletin-paseo/src/lib.rs
+++ b/runtimes/bulletin-paseo/src/lib.rs
@@ -135,21 +135,9 @@ pub mod migrations {
 	use super::*;
 
 	/// Unreleased migrations. Add new ones here:
-	pub type Unreleased = (
-		pallet_collator_selection::migration::v2::MigrationToV2<Runtime>,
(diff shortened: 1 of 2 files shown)


karolk91 (approved):

bkontur:
Old Paseo needs runtime upgrade to the latest: https://github.com/paritytech/polkadot-bulletin-chain/issues/612

Issue #612, opened 2026-06-17
[Testnet] Bulletin Paseo redeploy to latest


bkontur:
## TODO

- [x] runtimes-matrix.json return to `bulletin-paseo` in  `{ "name": "paseo",       "uris": ["wss://paseo-bulletin-rpc.polkadot.io"] },`
- [x] the latest runtime upgrade to old Paseo https://polkadot.js.org/apps/?rpc=wss://paseo-bulletin-rpc.polkadot.io (watch out, this is not Paseo Next V2)  (find sudo)

dr333ws:
Completed

bkontur:
thank you :)

PR #618, opened 2026-06-22
Add Paseo instance to try-runtime matrix

x3c41a:
Adds the original Bulletin Paseo (`wss://paseo-bulletin-rpc.polkadot.io`) as a try-runtime instance under `bulletin-paseo`, ahead of migrating it onto the `bulletin-paseo` runtime.

Note: this is *not* Paseo Next V2.

### Why under `bulletin-paseo` (not `bulletin-westend`)

old Paseo currently runs `bulletin-westend` (`1_000_015`), but every pallet storage version is already current except transaction-storage (v4 vs v5):

```
                     old-Paseo   PaseoNextV2
  ParachainSystem        3            3
  CollatorSelection      2            2
  XcmpQueue              6            6
  Session                1            1
  AuraExt                1            1
  TransactionStorage     4            5   <- only delta
  PolkadotXcm            1            1
```

The 1_000_015 `bulletin-westend` deploy already ran the historical migrations (ParachainSystem 2->3 etc.), so the earlier "ParachainSystem 2 vs 3" try-runtime failure no longer applies. `bulletin-paseo` wires `MigrateV4ToV5`, which closes the only remaining gap. So old Paseo can be upgraded directly onto `bulletin-paseo`, and its try-runtime check belongs under `bulletin-paseo` (with `--disable-spec-name-check`).

Refs #612

PR #619, opened 2026-06-22
Wire transaction-storage v3→v5 migrations for bulletin-paseo

x3c41a:
The live Bulletin Paseo chain is at transaction-storage storage version 3, while the in-code version is 5. The `bulletin-paseo` runtime wired no versioned migrations, so a runtime upgrade left the chain at v3 and failed the storage-version check.

Wires the staged path:
- `MigrateV3ToV4` (MBM) walks `AutoRenewals` to the v4 layout
- `MigrateV4ToV5` (single-block) translates the authorizer budget

Both are self-guarded and idempotent, so they no-op on Paseo Next V2 and previewnet, which are already at v5.

Note on sequencing: the two steps run at different times. The single-block v4→v5 fires during `on_runtime_upgrade` (skips while the chain is at v3); the MBM then walks v3→v4. So one upgrade moves an old v3 chain to v4, and a following upgrade takes it v4→v5.

Prerequisite for #612.

### DIFF
--- a/runtimes/bulletin-paseo/src/lib.rs
+++ b/runtimes/bulletin-paseo/src/lib.rs
@@ -136,8 +136,11 @@ pub mod migrations {
 
 	/// Unreleased migrations. Add new ones here:
 	///
-	/// Paseo is a fresh chain, so the historical version-bump migrations do not apply.
-	pub type Unreleased = ();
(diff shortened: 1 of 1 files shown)


bkontur (approved):

bkontur:
@x3c41a you could do it as a part or https://github.com/paritytech/polkadot-bulletin-chain/pull/618

PR #620, opened 2026-06-23
Fix shebang position in upgrade_runtime.js

x3c41a:
The shebang `#!/usr/bin/env node` was on line 4, after the license header. A shebang is only honored on line 1; mid-file Node parses `#` as code, so the script died with `SyntaxError: Invalid or unexpected token` and couldn't run at all.

Moved it to line 1, license header below it.

Verified the fixed script runs:
```
node upgrade_runtime.js --verify-only --network paseo
```

### DIFF
--- a/examples/upgrade_runtime.js
+++ b/examples/upgrade_runtime.js
@@ -1,7 +1,7 @@
+#!/usr/bin/env node
 // Copyright (C) Parity Technologies (UK) Ltd.
 // SPDX-License-Identifier: Apache-2.0
 
-#!/usr/bin/env node
 /**
  * Runtime Upgrade Script for Bulletin Chain Networks
  *


bkontur (approved):

examples/upgrade_runtime.js
```diff
@@ -1,7 +1,7 @@
+#!/usr/bin/env node
 // Copyright (C) Parity Technologies (UK) Ltd.
 // SPDX-License-Identifier: Apache-2.0
 
-#!/usr/bin/env node
```
bkontur, line 4: @x3c41a where do we use this?

x3c41a: We use it in our manual runtime-upgrade tool:
- https://github.com/paritytech/polkadot-bulletin-chain/blob/main/docs/playbook.md#L71
- https://github.com/paritytech/polkadot-bulletin-chain/blob/main/.claude/skills/release/skill.md#L47-L56

switch to `Code view` to see the exact lines

PR #621, opened 2026-06-23
bulletin-westend: bump spec_version to 1_000_016

x3c41a:
Bumps `bulletin-westend` spec_version 1_000_015 -> 1_000_016. No other runtime changes.

This is needed to enact the already-wired single-block `MigrateV4ToV5` transaction-storage migration. That migration only runs during `on_runtime_upgrade` while the chain is at storage v4, so it needs a fresh spec_version to trigger.

Context: old Paseo (which runs the `bulletin-westend` runtime) was just upgraded to v1_000_015 and walked storage v3 -> v4 via the MBM. The single-block v4 -> v5 step skipped that pass (chain was still at v3 at the upgrade block). Deploying this bump runs v4 -> v5 and brings it to storage v5.

No-op where already at v5 (the migration is version-guarded).

### DIFF
--- a/runtimes/bulletin-westend/src/lib.rs
+++ b/runtimes/bulletin-westend/src/lib.rs
@@ -182,7 +182,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: alloc::borrow::Cow::Borrowed("bulletin-westend"),
 	impl_name: alloc::borrow::Cow::Borrowed("bulletin-westend"),
 	authoring_version: 1,
-	spec_version: 1_000_015,
+	spec_version: 1_000_016,
 	impl_version: 1,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 1,


runtimes/bulletin-westend/src/lib.rs
```diff
@@ -182,7 +182,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: alloc::borrow::Cow::Borrowed("bulletin-westend"),
 	impl_name: alloc::borrow::Cow::Borrowed("bulletin-westend"),
 	authoring_version: 1,
-	spec_version: 1_000_015,
+	spec_version: 1_000_016,
```
bkontur, line 185: @x3c41a we should also upgrade bulletin-westend

bkontur (approved):

bkontur:
@karolk91 @x3c41a and also we should upgrade bulletin paseo with bulletin-paseo runtime instead of bulletin-westend :)

PR #640, opened 2026-06-26
Bump col11 eviction probe timeout to 450s

x3c41a:
`parachain_auto_renew_many_items_prune_eviction_test` flaked on `westend/mixed`: the first-item col11 eviction probe timed out at 300s while async cleanup lagged the finalized-height crossing under CI load. It passed in the other 7/8 matrix cells.

Bumps `BITSWAP_EVICTION_TIMEOUT_SECS` 300 -> 450. No logic change; fast-cleanup runs still return early.


bkontur (approved):

bkontur:
cc: @karolk91

karolk91:
As far as I'm aware, flaky tests are not related to timeouts - the only issue I'm aware of is being fixed by https://github.com/paritytech/polkadot-sdk/pull/12106 so we will see after its merged

karolk91 (changes requested):
@x3c41a as mentioned in the comment above, I did a re-run of CI and you can see some tests are still flaky - so the timeout increase will not help here. I would drop the change as there is no reason to increase it

Issue #458, opened 2026-04-27
[VIBE DRAFT - do not implement] console-cli: command-line interface for Polkadot Bulletin Chain


karolk91:
# console-cli: command-line interface for Polkadot Bulletin Chain

  ## Summary

  Build `console-cli` — a CLI tool that exposes every workflow `console-ui` offers (store, download, renew, authorize, query, faucet, status) — backed by `@parity/bulletin-sdk`. Starting point is the CLI half of #135 (with the web frontend stripped out). Scope, decisions, and risks below were vetted in a plan-review pass.

  ## Goals

  - Cover every workflow `console-ui` supports, scriptable and headless.
  - Reuse `@parity/bulletin-sdk` as the single source of truth — no duplicated chain logic.
  - Same network configs as `console-ui` (Westend, Paseo, local, previewnet).
  - Two output modes: human-readable tables and `--output json` / NDJSON for piping.

  ## Non-goals

  - No reactive state (RxJS) — CLI is request/response.
  - No browser wallet extensions.
  - No persistent history UI — defer to shell history + JSON output.

  ## What already exists (reuse, don't rebuild)

  | Asset | Reuse plan |
  |---|---|
  | #135 `examples/bulletin-helia/src/cli.ts` | Promote to `console-cli/src/commands/download.ts`, wrap in commander |
  | #135 `src/ipfs.ts` (319 lines) | Kept as Helia retrieval impl behind a new `RetrievalClient` interface |
  | #135 `logger-base.ts` + `logger-cli.ts` | Adopted; replaces ora/chalk |
  | #135 `tsconfig.cli.json` | Promoted to `tsconfig.json` |
  | `examples/authorize_and_store_papi*.js`, `store_*.js`, `cid_dag_metadata.js`, `native_ipfs_dag_pb_chunked_data.js`, `typescript/authorize_and_store.js` | Logic patterns reused; scripts  retired post-launch |
  | `console-ui/src/config/networks.ts` | Copied into CLI (decision: copy over shared package — see below) |
  | `console-ui/src/state/storage.state.ts` read patterns | Reference for `query` commands |
  | `@parity/bulletin-sdk` | Sole interface to chain ops; CLI is a thin shell |
  | `examples/authorize_and_store_papi_smoldot.js` | Preserved via `--smoldot` global flag |

  #135's frontend half (`index.html`, `main.ts`, `style.css`, `logger.ts`, web `tsconfig.json`, vite deps) is dropped.

[...]

PR #642, opened 2026-06-29
utils: IPFS gateway debug tools

x3c41a:
Adds a few small tools under `utils/ipfs-debug/` for tracing a missing IPFS CID on the Bulletin gateway.

## Why

While investigating a 504 storm on the Paseo bulletin gateway, we traced it to Polkadot Desktop polling a single CID once a minute for hours, every request timing out because no node had the content. These tools answer the two questions that came up: is the CID actually on the chain, and (for web content) which app references it.

## What's included

- `bulletin-has-cid.mjs` — the authoritative check. `store` content-addresses with blake2b + raw codec, so a `bafk2bzace…` CID maps to an on-chain `content_hash`. This queries `TransactionByContentHash` to tell whether the CID is stored (within retention) or gone.
- `find_cid_parent.py` — given dag-pb roots, walks each on the gateway and reports whether a target CID is a descendant, and at what path. Matches on the parent's links, so it works even though the target itself 404/504s.
- `find-app-for-cid.sh` — resolves candidate `.dot` names to bundle roots via the `dotns` CLI, then walks each with the script above.
- `README.md` — usage and a note on hash types (dotNS web bundles use sha256, Bulletin-native content uses blake2b).

## Notes

- The Python tools are dependency-light (need `curl`). `bulletin-has-cid.mjs` needs a papi project with the `bulletin` descriptor.
- All three were tested against live Paseo: resolving real apps, walking real bundles on the gateway, and a chain lookup confirming the offending CID is not stored.

### DIFF
--- a/utils/ipfs-debug/README.md
+++ b/utils/ipfs-debug/README.md
@@ -1,0 +1,60 @@
+# IPFS gateway debug tools
+
+Small tools for tracing a missing IPFS CID: deciding whether it is on the
+Bulletin chain at all, and (for web content) which dotNS app references it.
+Built while investigating a 504 storm on the Paseo bulletin gateway, where
+Polkadot Desktop polled one missing CID once a minute for hours.
+
+## `bulletin-has-cid.mjs` — is this CID stored on Bulletin?
+
+The authoritative check. `pallet-transaction-storage::store` content-addresses
+with blake2b + raw codec, so a `bafk2bzace…` CID maps directly to an on-chain
+`content_hash`. This queries `TransactionByContentHash` for that digest.
+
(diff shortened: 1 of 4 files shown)


karolk91:
I dont like these are separate scripts, I see such utils as part of some general CLI utli: https://github.com/paritytech/polkadot-bulletin-chain/issues/458 - so please add some ideas to the thread there about features it could support instead of these separate utils

bkontur:
> I dont like these are separate scripts, I see such utils as part of some general CLI utli: #458 

wen "bulletin-cli" :smile:  :smile: ?

PR #653, opened 2026-07-14
scripts: verify operator-doc flags against pinned SDK source

x3c41a:
Extracts the Bulletin/HOP CLI flags from polkadot-sdk at the pinned node commit and checks a node-operator doc for renamed, removed, or missing flags.


karolk91:
I'm not sure if we need this changes - I would rather improve our CI-based HOP tests to fail if any of the params change in incompatible way

dr333ws:
> I'm not sure if we need this changes - I would rather improve our CI-based HOP tests to fail if any of the params change in incompatible way

that was a part of HOP flags update task. I posted my approach to github for verifiability

PR #661, opened 2026-07-15
Added check-runtime step to bulletin CI

antkve:


karolk91:
@antkve - could you check what is to be done here to push this forward ?

PR #659, opened 2026-07-15
feat: add Bulletin paseo

alexbcasas:
I am adding Bulletin Paseo to this portal after onboarding Bulletin chain as a system chain in Paseo.

PR #663, opened 2026-07-15
removed paseo uri which is down

rosarp:


scripts/runtimes-matrix.json
```diff
@@ -24,7 +24,6 @@
       "spec_name_check": "--disable-spec-name-check",
       "extra_flags": "--blocktime 24000 --disable-spec-version-check",
       "instances": [
-        { "name": "paseo",      "uris": ["wss://paseo-bulletin-rpc.polkadot.io"] },
```
bkontur, line 27: Check Also console-ui: https://github.com/paritytech/polkadot-bulletin-chain/blob/601ffbe988f0e2f2232417ac69d4e4e9b5abb657/console-ui/src/config/networks.ts#L170

bkontur (approved):

x3c41a (approved):

karolk91:
I suggest instead of removing - switch to the one proposed by https://github.com/paritytech/polkadot-bulletin-chain/pull/659

But also make sure to change this all over the affected places (console ui, chainspecs, docs, examples) - also generate proper PAPI descriptors where these are used for new endpoint
 
cc @AlexD10S

dr333ws:
For the record: `paseo-bulletin-rpc.polkadot.io` does not resolve anymore (DNS record is gone, the sibling endpoints still resolve), so every migration run with this instance dies in try-runtime with `Input("At least one client must be provided")` - same failure on main since this morning.

karolk91 (approved):

Issue #579, opened 2026-05-29
[Testing] Use `bitswap_v1_get` node RPC as additional check besides IPFS retrievals


bkontur:
Now all the `examples` and zombienet sdk tests using Bitswap  (IPFS, Helia, ..), but we should also check content over new RPC for bitswap from node directly.

For example:
- we have `fetchCid`, we could add `fetchCidFromNode`
- add `fetchContent(cid, httpIpfsApi, nodeWsUrl) which will fetch from both and compare that it is 100% sure and returns data content`
- replace fetchCid with fetchContent in tests

karolk91:
lets also add Bitswap RPC as another tab in the Download section to console-ui

PR #665, opened 2026-07-16
ci: cover both rocksdb and paritydb database backends

x3c41a:
## Why

CI never set `--database`, so every test network ran the omni-node default, which turns out to be RocksDB (verified in sc-cli source and with the pinned binaries). ParityDb, the backend production currently runs, had zero coverage. The backend only affects the node-local storage path (indexed transactions, pruning, sync, bitswap serving) and only changes via node version bumps, so coverage is scoped to where it matters.

## What

- Integration tests run a mixed-backend network: collator-1 rocksdb, collator-2 paritydb, set directly in the zombienet TOMLs. The authorize-and-store test also verifies the stored transaction on collator-2. No extra CI legs.
- zombienet-sdk suites accept a `BULLETIN_DB_BACKEND` env var (`rocksdb` | `paritydb`); unset keeps today's behavior.
- New paritydb legs, westend only: the sync suite (behind its existing dispatch/label gate) and the auto-renew pruning/restart/mixed groups. The auto-renew legs run on main pushes, dispatch, and PRs touching DB-relevant files (new `db-files-changed` output in check-changed-files); ordinary PRs skip them.
- HOP suite and the archive group are unchanged, the backend can't affect them.

Cost: ordinary PRs unchanged; main pushes and DB-relevant PRs get +3 parallel legs; sync runs get +1.

Verified locally: the mixed network spawns with RocksDb/ParityDb per the collator logs, authorize-and-store passes including the collator-2 check, and the fast-sync sdk test passes fully on paritydb.


zombienet/bulletin-paseo-local.toml
```diff
@@ -45,6 +45,7 @@ validator = true
 p2p_port = 10001
 rpc_port = 10000
 args = [
+	"--database {{BULLETIN_DB_BACKEND}}",
```
bkontur: @karolk91 @x3c41a do we need this BULLETIN_DB_BACKEND env? Why not just set it up here directly?

```suggestion
	"--database rocksdb",
```

x3c41a: Nah, the env was was not doing anything useful here; removed it. Now both collators hardcode their backend. 

The env var stays only for the zombienet-sdk tests.

dr333ws:
prepare-snapshot (bulletin-paseo) fails all 10 attempts with remote-externalities `Input("At least one client must be provided")`, i.e. the paseo RPC endpoint never accepted a connection; unrelated to the rocksdb/paritydb change. Resolution: re-run the job once the endpoint is healthy (or point the workflow at a working paseo RPC).

examples/authorize_and_store_papi.js
```diff
@@ -5,25 +5,28 @@ import assert from "assert";
...
 import { cryptoWaitReady } from '@polkadot/util-crypto';
-import { authorizeAccount, fetchCid, store, TX_MODE_FINALIZED_BLOCK } from './api.js';
+import { authorizeAccount, fetchCid, store, verifyStoredOnNode, TX_MODE_FINALIZED_BLOCK } from './api.js';
 import { setupKeyringAndSigners, waitForBlockProduction, DEFAULT_IPFS_GATEWAY_URL } from './common.js';
 import { logHeader, logConnection, logSuccess, logError, logTestResult } from './logger.js';
 import { cidFromBytes } from "./cid_dag_metadata.js";
 import { bulletin } from './.papi/descriptors/dist/index.js';
 
-// Command line arguments: [ws_url] [seed] [ipfs_api_url]
+// Command line arguments: [ws_url] [seed] [ipfs_api_url] [second_node_ws]
```
karolk91: Instead of `[second_node_ws]` I would probably allow to specify multiple urls via `[ws_url]` and just iterate in a loop

x3c41a: made it a list

examples/api.js
```diff
@@ -220,6 +220,31 @@ export async function fetchCid(httpIpfsApi, cid) {
     return Buffer.from(await res.arrayBuffer())
 }
 
+/**
+ * Verify a specific node imported and indexed a stored transaction: its finalized
+ * head reached `blockNumber` and the on-chain metadata at that block contains the
+ * CID's content hash. Used to assert on both collators of the mixed-backend test
+ * network (collator-1 rocksdb, collator-2 paritydb).
+ */
+export async function verifyStoredOnNode(client, typedApi, blockNumber, cid, timeoutSec = 120) {
```
karolk91, line 229: I dont see a lot of value in checking if the block was propagated etc - no harm to do it but if there is a problem with that it would point to a whole different class of fundamental errors (i.e. its fundamental that different nodes reach consensus in finalized state - we take it for granted here)

I would rather want we check if the CID can be downloaded directly from each node that would prove its stored properly in database. 

Please check my [PR](https://github.com/paritytech/polkadot-bulletin-chain/pull/135) that introduces small lib that allows to read data directly from specific node - but its not using the websocket url and rather P2P multiaddresses of the nodes. We can merge it and use it as explained in the PR description (see Library usage in the PR description)

bkontur: > I would rather want we check if the CID can be downloaded directly from each node that would prove its stored properly in database.
 

agree, when we have a list of nodes now, I would just extend the fetchCid check to pass all over the node list and call the Bitswap RPC on the node and cover/resolve another issue with that: https://github.com/paritytech/polkadot-bulletin-chain/issues/579

karolk91: I was thinking about straight bitswap over P2P (not to confuse with RPC `bitswap_v1_get`). But I agree we can go with RPC `bitswap_v1_get`

@x3c41a please work on https://github.com/paritytech/polkadot-bulletin-chain/issues/579 and push as separate PR, then we will switch this implementation to use that

bkontur: > I was thinking about straight bitswap over P2P

I am also fine with this as another check.

Not sure if "Node RPC" and "Bitswap over P2P" exercise the same DB columns. But more coverage we have, then better, also in context  in rocksdb vs paritydb exercises. 


Ideally, my idea was to have one `fetchCid` (or something) for examples, which returns just one `data` and internally it trigger and comapre the data across all the supported ways:
- all the nodes list Bitswap RPC 
- IPFS gateway way
- smoldot Bitswap RPC
- Karol's client way

it would mean no changes for tests structure and we exercise all the ways in one place

x3c41a: #579 implemented in #668 

fetchContent now cross-checks gateway vs bitswap_v1_get in all the examples so each node in the list must now serve the stored bytes from its own DB.  Final tests are still running...

Also added a Bitswap RPC to console-ui.

x3c41a: https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/29576440225

x3c41a: flaky eviction test again (same as yesterday, this time on paseo ), not related to this PR, re-running

karolk91 (commented):
Please sync with latest changes from https://github.com/paritytech/polkadot-bulletin-chain/pull/669

dr333ws:
@bkontur could you take a look at the PR while @karolk91 is away

PR #668, opened 2026-07-17
testing: cross-check content via bitswap_v1_get node RPC

### LINKED ISSUE #579: [Testing] Use `bitswap_v1_get` node RPC as additional check besides IPFS retrievals

x3c41a:
Closes #579. Stacked on #665 (retargets to main once that merges); implements the direction from the review thread there: verification should download the data from each node, not only check state.

- `fetchCidFromNode` fetches a stored block via the node's `bitswap_v1_get` RPC; `fetchContent` fetches from the IPFS gateway and the node RPC and asserts both return identical bytes. All example retrieval paths now go through `fetchContent`.
- `verifyStoredOnNode` additionally fetches the data back from each node in the `ws_url` list, so both collators of the mixed-backend network (rocksdb and paritydb) must serve the bytes from their own database.
- The smoldot example takes an optional node WS URL for the cross-check (a light client cannot serve custom RPCs).
- zombienet-sdk `verify_node_bitswap` now cross-checks over the RPC as well, covering p2p bitswap and RPC in one call.
- console-ui gets a "Bitswap RPC" download tab backed by the same method.

Verified locally against the mixed-backend network (collator-1 rocksdb, collator-2 paritydb): authorize-and-store ws (per-node retrieval on both backends) and smoldot, chunked, preimage, dag-pb and big-data flows, console-ui build + e2e (7 passed), and the sdk fast-sync test with the RPC cross-check. Full CI kicks in when #665 merges and this retargets to main.


dr333ws:
Screenshot 2026-07-17 at 13 27 15

examples/authorize_and_store_papi_smoldot.js
```diff
@@ -109,22 +110,26 @@ async function main() {
...
+        console.error('  For solochains: <solo-chain-spec-path> [ipfs-api-url] [node-ws-url]');
         process.exit(1);
     }
 
     // Optional parachain chainspec path (only needed for parachains)
     const parachainSpecPath = process.argv[3] || null;
     // Optional IPFS API URL
     const HTTP_IPFS_API = process.argv[4] || DEFAULT_IPFS_GATEWAY_URL;
+    // Optional node WS URL for the bitswap_v1_get cross-check (smoldot cannot
+    // serve custom RPC methods, so this needs a direct node connection).
```
karolk91: smoldot also supports bitswap RPC

x3c41a: Right, fixed!

The smoldot test now calls bitswap_v1_get on the smoldot client itself, no extra node connection. Verified on a local zombienet.

karolk91:
> Closes https://github.com/paritytech/polkadot-bulletin-chain/issues/579. Stacked on https://github.com/paritytech/polkadot-bulletin-chain/pull/665 (retargets to main once that merges);

I think it should be the other way around - https://github.com/paritytech/polkadot-bulletin-chain/pull/665 can be stacked on top of this PR

@x3c41a, lets target `master` branch with this PR (we will then have all the CIs run).

dr333ws:
Reopened against `main` as #669

PR #669, opened 2026-07-20
testing: cross-check content via bitswap_v1_get node RPC

### LINKED ISSUE #579: [Testing] Use `bitswap_v1_get` node RPC as additional check besides IPFS retrievals

x3c41a:
Closes #579.

Adds a `bitswap_v1_get` node-RPC check alongside the existing IPFS retrievals, so stored content is verified over both paths.

- `fetchCidFromNode` fetches a stored block via the node's `bitswap_v1_get` RPC; `fetchContent` fetches from the IPFS gateway and the node RPC and asserts both return identical bytes. All example single-block retrieval paths now go through `fetchContent` (gateway-only `fetchCid` stays where the gateway assembles a dag-pb DAG from a root CID).
- The smoldot example uses its own `bitswap_v1_get` for the cross-check (smoldot forwards the request to its peers over p2p bitswap).
- zombienet-sdk `verify_node_bitswap` now cross-checks over the RPC as well, covering p2p bitswap and RPC in one call.
- console-ui gets a "Bitswap RPC" download tab backed by the same method.

Supersedes #668, which was stacked on #665; this targets `main` directly so full CI runs. The per-node `verifyStoredOnNode` download check stays in #665.

Verified locally against a zombienet network: authorize-and-store (ws + smoldot), chunked, preimage, dag-pb and big-data flows, and the console-ui build + Bitswap RPC tab driven end to end.


console-ui/src/pages/Download/Download.tsx
```diff
@@ -445,6 +449,14 @@ export function Download() {
             // not valid JSON despite content-type
           }
         }
+      } else if (activeTab === "rpc" && client) {
+        data = await fetchFromBitswapRpc(client, parsedCid.toString());
+        try {
+          parsedJSON = JSON.parse(new TextDecoder().decode(data));
+          isJSON = true;
+        } catch {
+          // not JSON content
```
x3c41a, line 458: anything to catch here?

examples/api.js
```diff
@@ -220,6 +220,35 @@ export async function fetchCid(httpIpfsApi, cid) {
...
+ *
+ * Only valid for CIDs of single stored blocks: for dag-pb root CIDs the gateway
+ * returns the assembled file while the node returns the indexed block.
+ */
+export async function fetchContent(cid, httpIpfsApi, client) {
+    const [fromIpfs, fromNode] = await Promise.all([
+        fetchCid(httpIpfsApi, cid),
+        fetchCidFromNode(client, cid),
+    ]);
+    assert(
```
karolk91: I think it makes more sense to verify the data against CID itself (CID consists of data algorithm and data hash) so we can hash the returned data accordingly and compare with CID encoded info (see dag-pb info below)

It also make sense to me to centralize this verification logic here instead of every test script checking this on its own.

We also need to take into account the cases for `raw` codec and `dag-pb` codec. Since CI tests seems to pass properly I think it proves we don't have proper `dag-pb` related testing because this function would fail - IPFS gateway works in a way that it will "compose" all the child CIDs related to `dag-pb` root item and return everything as one response, while Bitswap RPC would only return that root item itself - so both responses would not match.

Try to improve this across tests:
1. verify logic based on CIDs in single place (here)
2. verify dag-pb case test scenario - imho it should fail given we have this function but it doesn't fail
3. hash of `fromIpfs` returned data in `dag-pb` scenario will not match the CID - so we would likely need `fromIpfsRaw` verision that uses `format=raw` in the HTTP query and compare this against CID

console-ui/src/pages/Download/Download.tsx
karolk91: I'm not sure but seems to me like its just "best effort" logic to try if the response is json - but of course it can be anything so we can't assume its json

examples/api.js
x3c41a: - added verifyCid that hashes the block against the CID digest, 
- fetchContent now pulls the raw block from both the gateway and bitswap and verifies each against the CID, 
- and the dag-pb test runs the real (0x70) root CID through it so that path is actually covered now

Verified locally, now waiting for CI to confirm local run

console-ui/src/pages/Download/Download.tsx
x3c41a: that's my understanding as well after re-reading the code with Claude

examples/api.js
```diff
@@ -220,6 +220,67 @@ export async function fetchCid(httpIpfsApi, cid) {
...
+    verifyCid(fromIpfs, cid);
+    verifyCid(fromNode, cid);
+    assert(
+        fromIpfs.equals(fromNode),
+        `❌ IPFS gateway and node RPC content mismatch for CID ${cid.toString()} ` +
+        `(${fromIpfs.length} vs ${fromNode.length} bytes)`,
+    );
+    return fromIpfs;
+}
+
```
karolk91, line 285:
```suggestion
/** A retrieval path: a name plus an async `(cid) => Buffer` fetcher of the raw block. */
  export const gatewaySource = (httpIpfsApi) =>
      ({ name: `ipfs-gateway ${httpIpfsApi}`, fetch: (cid) => fetchCidRaw(httpIpfsApi, cid) });

  export const nodeRpcSource = (client, label = 'bitswap-rpc') =>
      ({ name: label, fetch: (cid) => fetchCidFromNode(client, cid) });

  /**
   * Fetch the block `cid` addresses from every source and assert each hashes to the CID.
   * Verifying every block against the CID also proves they are byte-identical (equal
   * digests under a collision-resistant hash), so no separate cross-source compare is
   * needed. Returns the block — the raw node for dag-pb CIDs, not the assembled file.
   */
  export async function fetchAndVerifyBlockFrom(cid, sources) {
      assert(sources.length >= 2, `cross-check needs ≥2 sources, got ${sources.length}`);
      const blocks = await Promise.all(
          sources.map(async (s) => ({ name: s.name, data: await s.fetch(cid) })),
      );
      for (const { name, data } of blocks) verifyCid(data, cid, name);
      return blocks[0].data;
  }

```

we could then call something like below in the test, wdyt?:

```
 const block = await fetchAndVerifyBlockFrom(cid, [
      gatewaySource(HTTP_IPFS_API),
      nodeRpcSource(collator1Client, 'collator-1 rocksdb'),
      nodeRpcSource(collator2Client, 'collator-2 paritydb'),
  ]);
```

x3c41a: sounds good but I just used a rest param (cid, ...sources) instead of a wrapped array. 
I thought about pulling the gateway out or making a struct, but keeping all sources the same is simpler since each block is checked against the CID anyway. also gave verifyCid a label arg.

karolk91 (approved):

examples/authorize_and_store_papi_smoldot.js
```diff
@@ -162,8 +162,9 @@ async function main() {
         const { cid } = await store(bulletinAPI, whoSigner, dataToStore);
         logSuccess(`Data stored successfully with CID: ${cid}`);
 
-        // Read back from IPFS
-        let downloadedContent = await fetchCid(HTTP_IPFS_API, cid);
+        // Read back from IPFS and smoldot's bitswap_v1_get (forwarded to peers
+        // over p2p bitswap), verifying both match.
+        let downloadedContent = await fetchAndVerifyBlock(cid, gatewaySource(HTTP_IPFS_API), nodeRpcSource(client));
```
bkontur, line 167: @x3c41a we have here also smoldot RPC, can you please add follow-up with another source: `smoldotRpcSource(..)`?

Otherwise PR looks good 🚀

PR #670, opened 2026-07-21
testing: add smoldotRpcSource and node RPC cross-check to smoldot example

x3c41a:
Follow-up to [this comment](https://github.com/paritytech/polkadot-bulletin-chain/pull/669#discussion_r3620704582) on #669.

- `smoldotRpcSource(client)` names smoldot's `bitswap_v1_get` retrieval path (the smoldot example already used it, but under the generic `nodeRpcSource` label).
- The smoldot example takes an optional node WS URL and, when given, also cross-checks the node's `bitswap_v1_get` RPC, so the stored block is verified over three paths: IPFS gateway, smoldot p2p bitswap and node RPC.
- The justfile passes the node WS URL in both smoldot invocations, so CI and local runs cover all three.

Verified locally against a zombienet network (kubo-native): the smoldot example fetched the block from the gateway, smoldot and the node RPC, and all hashed back to the CID.

### DIFF
--- a/examples/api.js
+++ b/examples/api.js
@@ -267,6 +267,9 @@ export const gatewaySource = (httpIpfsApi) =>
 export const nodeRpcSource = (client, label = 'bitswap-rpc') =>
     ({ name: label, fetch: (cid) => fetchCidFromNode(client, cid) });
 
+/** Smoldot serves the same `bitswap_v1_get`, fetching the block from its peers over p2p bitswap. */
+export const smoldotRpcSource = (client) => nodeRpcSource(client, 'smoldot-bitswap-rpc');
+
 /**
  * Fetch the block `cid` addresses from every source and assert each hashes to the
  * CID. Verifying every block against the CID also proves they are byte-identical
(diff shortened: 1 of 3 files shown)


examples/api.js
```diff
@@ -267,6 +267,9 @@ export const gatewaySource = (httpIpfsApi) =>
 export const nodeRpcSource = (client, label = 'bitswap-rpc') =>
     ({ name: label, fetch: (cid) => fetchCidFromNode(client, cid) });
 
+/** Smoldot serves the same `bitswap_v1_get`, fetching the block from its peers over p2p bitswap. */
+export const smoldotRpcSource = (client) => nodeRpcSource(client, 'smoldot-bitswap-rpc');
```
bkontur, line 271: wow, so easy? 🚀

bkontur (approved):

examples/api.js
bkontur: haha, right, it was working before, but I didnt realize that :D, client was smoldot one :D

```
let downloadedContent = await fetchAndVerifyBlock(cid, gatewaySource(HTTP_IPFS_API), nodeRpcSource(client));
```

PR #674, opened 2026-07-22
examples: fix auto-renew recipe script name

x3c41a:
Rename `check_auto_renew_papi.js` to `check_auto_renew_chopsticks.js` (it drives chopsticks; the papi name was an oversight in #313), which makes the previously broken `run-test-auto-renew-chopsticks` recipe work unchanged.


examples/justfile
```diff
@@ -607,7 +607,7 @@ run-test-auto-renew-chopsticks endpoint="wss://westend-bulletin-rpc.polkadot.io"
         echo "✅ WASM built: $WASM_PATH"
     fi
 
-    node check_auto_renew_chopsticks.js "{{ endpoint }}" "$WASM_PATH"
+    node check_auto_renew_papi.js "{{ endpoint }}" "$WASM_PATH"
```
rosarp: instead of this, rename file.
https://github.com/paritytech/polkadot-bulletin-chain/blob/main/examples/check_auto_renew_papi.js
is actually using chopsticks! It was my oversight.. :)

rosarp: Will do this in another PR.. let this sync in right now.

rosarp (approved):

examples/justfile
x3c41a: renamed the file and reverted the recipe line, so the diff is now just the rename

PR #676, opened 2026-07-22
testing: cover untested migrations (v4->v5, retention-period reset, isolated v1->v2)

x3c41a:
Add unit tests for `MigrateV4ToV5` (budget translation + provider refs), `SetRetentionPeriodIfZero`, and an isolated `MigrateV1ToV2` transformation check; all three still run on live westend, which is at on-chain storage version 0.


pallets/transaction-storage/src/tests.rs
```diff
@@ -1474,6 +1474,103 @@ fn migration_v1_empty_storage() {
...
+		RetentionPeriod::set(0);
+
+		SetRetentionPeriodIfZero::<Test, ConstU64<77>>::on_runtime_upgrade();
+
+		SetRetentionPeriodIfZero::<Test, ConstU64<77>>::post_upgrade(Vec::new())
+			.expect("retention period is non-zero after migration");
+	});
+}
+
+// ---- v1 → v2 migration tests ----
```
x3c41a, line 1521: the question is whether we still need v1->v2 migration tests or we only need the latest v4->v5?

x3c41a: Okay, checked myself and live bulletin westend  is still at on-chain storage version 0, so the next upgrade there actually runs v1→v2. The test can be dropped once westend is past v2 and the old migrations are dropped from the runtime.

PR #678, opened 2026-07-22
testing: validate genesis config presets for both runtimes

x3c41a:
Build every named genesis preset of both runtimes into real storage and assert sudo key, authorizers, para id, endowments, invulnerables, and session keys.


dr333ws:
closing to keep the review queue focused: presets are exercised by every zombienet spec build anyway; branch stays around, tracked in #687 if we want it later

PR #680, opened 2026-07-22
ci: run zombienet sync suite weekly

x3c41a:
Run the zombienet sync suite weekly (Mon 03:00 UTC) in addition to dispatch/label; other jobs stay off the schedule and concurrency groups are split per event so cron cannot cancel push runs.


bkontur:
@x3c41a we run this on every PR and when merged to main, what is the purpose/motivation run it also once per week? Do we expect any external changes every week or something?

dr333ws:
> @x3c41a we run this on every PR and when merged to main, what is the purpose/motivation run it also once per week? Do we expect any external changes every week or something?

It's the other way around for `sync` tests actually. The sync tests never run today (not on PRs, not when we merge main), only if someone starts them by hand or adds a label. Other zombienet tests (auto-renew, hop,...) do run as you mentioned but not `sync`.

Weekly runs will help us catch sync failures within a week. Otherwise, the failure is caught whenever someone tries to sync a node.

bkontur:
> Weekly runs will help us catch sync failures within a week. Otherwise, the failure is caught whenever someone tries to sync a node.

Doen't it make sense to run those nightly (run after midnight)? If we need to catch errors, why wait the while week. When they fail, do they report somehow? How do we see the error?

dr333ws:
> > Weekly runs will help us catch sync failures within a week. Otherwise, the failure is caught whenever someone tries to sync a node.
> 
> Doen't it make sense to run those nightly (run after midnight)? If we need to catch errors, why wait the while week. When they fail, do they report somehow? How do we see the error?

Yeah, we can run it daily too. I don't really know how often the related codebase changes, did not want to unnecessarily overload the CI.  

> When they fail, do they report somehow? How do we see the error? 

red runs + emails from GH

PR #682, opened 2026-07-22
Fix flaky vitest integration timeouts and stop hiding failures

### LINKED ISSUE #443: Introduce Optimal Submit algorithm to Bulletin SDK
Bulletin SDK should provide to its client the optimal way of working with regards to uploading data to it

The task is to follow the idea of optimal algorithm as outlined in: https://hackmd.io/@karolk/SJsuZNa3We as a default implementation of submitting to Bulletin

TODO: detailed description of goals and architecture within Bulletin SDK

x3c41a:
Use the production 420s `txTimeout` in vitest integration tests, derive vitest budgets from it, and drop `continue-on-error` so failures count again (fixes #443, supersedes #442).

Depends on #692 and #693 (SDK fixes; this PR's integration legs stay red until both are merged and the branch is refreshed).


franciscoaguirre (approved):
So just increasing the timeout? I don't mind

RafalMirowski1 (changes requested):
Removing `continue-on-error: true` made me notice that TypeScript SDK vitest integration tests fail immediately on `bulletin-paseo` with:
```
BulletinError: Failed to authorize account:
 InvalidTxError: {
  "type": "Invalid",
  "value": {
    "type": "BadSigner"
  }
}
```
My claude tells me this is because we sign with `//Alice` who is not an authorizer on paseo (but is on westend so tests don't fail there) and we can sign with `//Eve` instead, who is.

dr333ws:
@RafalMirowski1 good catch! fixed in the latest commit.
Now authorize is  signed by //Eve (the seeded authorizer), stores stay on //Alice

Please re-review

RafalMirowski1 (approved):

PR #684, opened 2026-07-22
benchmarking: add dedicated store_with_cid_config benchmark

x3c41a:
Add a dedicated `store_with_cid_config` benchmark using worst-case SHA2-256 (~30-37% slower per byte than blake2b, so the extrinsic was underweighted) and wire the extrinsic to its own weight; seed numbers to be regenerated with `/cmd bench`.


franciscoaguirre (approved):
Looks good, let's /cmd bench

bkontur:
@x3c41a nice, thank you

PR #677, opened 2026-07-22
testing: add unit tests for bulletin-pallets-common

x3c41a:
Add `pallets/common` unit tests for the utility call inspectors and `NoCurrency` fungible no-ops, on a minimal System+Utility mock.

PR #685, opened 2026-07-22
refactor: remove unused sudo/proxy call inspectors

x3c41a:
## What

- Delete `inspect_sudo_wrapper`, `inspect_proxy_wrapper`, `sudo_inner_calls` and `proxy_inner_calls` from `bulletin-pallets-common`; nothing in the repo calls them (both runtimes only inspect `Utility` wrappers and intentionally skip sudo, see `storage.rs`).
- Drop the now-unused `pallet-proxy` and `pallet-sudo` dependencies from the crate, and remove `pallet-proxy` from the workspace dependencies (no other crate uses it; the runtimes keep their own `pallet-sudo` dependency).
- Reword the `NoCurrency` doc comment to point at its actual consumer instead of `pallet-proxy`/`pallet-session`.

## Why

Minimal API surface: dead public functions invite accidental use and drag two pallet dependencies into every build of the crate.

Based on `main`; independent of #677 (after its latest trim commit the tests no longer reference the sudo/proxy inspectors, so the two merge cleanly in either order).

### DIFF
--- a/pallets/common/src/lib.rs
+++ b/pallets/common/src/lib.rs
@@ -40,8 +40,8 @@ use scale_info::TypeInfo;
 /// Fungible currency implementation that does not support any balance operations.
 /// Works only with zero balances.
 ///
-/// Note: This is a workaround to satisfy the `pallet-session::Config::Currency` and
-/// `pallet-proxy::Config::Currency` trait requirements.
+/// Note: This is a workaround to satisfy currency trait bounds (e.g.
+/// `pallet_bulletin_transaction_storage::Config::Currency`) in runtimes that hold no balances.
 pub struct NoCurrency<AccountId, HoldReason = ()>(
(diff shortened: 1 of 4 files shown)


pallets/common/src/lib.rs
```diff
@@ -387,36 +365,3 @@ pub fn utility_inner_calls<T: pallet_utility::Config>(
 		pallet_utility::Call::__Ignore(..) => vec![],
 	}
 }
-
-/// Extract inner calls from a proxy call variant.
-pub fn proxy_inner_calls<T: pallet_proxy::Config>(
```
bkontur, line 392: @x3c41a sudo can go away probably, but not sure about proxy, but we can add it later when we need this for BulletinPolkadot

bkontur (approved):

PR #688, opened 2026-07-23
sdk: keep in_block tx subscriptions alive until finalization

x3c41a:
Chunked uploads unsubscribed each chunk's PAPI submit observable at the first best-block sighting, which stops the node-side broadcast (transaction_v1_stop), so a reorged-out chunk was never re-included and the next chunk, already signed with the successor nonce, aged out of its mortality window and failed with AncientBirthBlock (CI failure unmasked by #682); the client now keeps the subscription and broadcast alive in the background until the transaction finalizes.


bkontur:
@RafalMirowski1 please, take a look here

RafalMirowski1 (approved):

PR #691, opened 2026-07-24
ci: split mixed zombienet group into three parallel cells

x3c41a:
Run the six standalone `mixed` scenarios as three parallel matrix cells (`mixed-1/2/3`, balanced so the 17-min worst-case test gets the lightest partner), cutting wall clock from ~56 to ~22 min with identical coverage; the hand-maintained `mixed` union group is removed.

### DIFF
--- a/justfile
+++ b/justfile
@@ -125,6 +125,15 @@ test-zombienet-auto-renew runtime="westend" group="all":
             filter_args=("{{group}}")
             ;;
     esac
+    # A filter matching zero tests exits 0; fail loudly instead (e.g. a stale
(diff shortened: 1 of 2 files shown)


justfile
bkontur: @x3c41a and remove this?

bkontur: or maybe just trigger `mixed-1)` 2 3?

x3c41a: removed in the last push.  `all` covers run-everything now. 
If anyone later wants a single local command for these six, we can re-add `mixed` as an alias for those 3 slices

bkontur (approved):
@x3c41a good job!

PR #693, opened 2026-07-24
sdk: retry once on mortality era expiry (AncientBirthBlock)

x3c41a:
The txpool can silently drop a broadcast transaction around a reorg and PAPI only reports it as InvalidTxError/AncientBirthBlock once the 64-block era boundary finalizes, at which point the original signature can never be included, so the SDK now re-signs and resubmits once (fresh mortality anchor and nonce) instead of failing the store.

### DIFF
--- a/sdk/typescript/src/async-client.ts
+++ b/sdk/typescript/src/async-client.ts
@@ -237,7 +237,6 @@ interface MappedTxStatus {
   }
 }
 
-/** Result of a completed sign-submit-watch cycle. */
 interface SignSubmitResult {
   blockHash: string
   txHash: string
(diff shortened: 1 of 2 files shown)


RafalMirowski1 (approved):
looks good, just thinking whether we retry should be configurable, wdyt?

dr333ws:
> looks good, just thinking whether we retry should be configurable, wdyt?

Mhm, generally good approach to have timeout configurable but in this case the benefit is questionable:
- the retry happens on the errors that are "safe"
- we don't want to turn the retry off
- an extra attempt costs another 64 blocks which is expensive 

So keeping as is, is rationale IMO

PR #662, opened 2026-07-15
Bump polkadot-sdk to stable2606

rosarp:
Bumps the workspace to the polkadot-sdk `stable2606` release line (from `stable2604`).
Closes #654.

## API breakage fixed
New in stable2606 (per `asset-hub-westend-runtime 0.49.0`):
- `cumulus_pallet_parachain_system::Config`: added `type SchedulingSignatureVerifier = ();`
- `RelayParentOffsetApi` (now api_version 2): implemented `max_claim_queue_offset()`.

## Node pin
- `POLKADOT_NODE_VERSION` moved to `73b734d929e97ac0700b13435bd3f85d1b720a25`
  (stable2606, HOP-enabled). It includes the sp-database refcount fix   ([#12106](https://github.com/paritytech/polkadot-sdk/pull/12106), backported as [#12678](https://github.com/paritytech/polkadot-sdk/pull/12678)), which aggregates
  same-key refcount ops within a DB transaction. This fixes `parachain_renew_twice_within_block_with_pruning_test`: two renews in one block previously leaked a `col11` refcount, so pruned data was still served over bitswap.

## Test fixes
- smoldot integration test (`authorize_and_store_papi_smoldot.js`): submit the authorize/store transactions with an immortal era. The mortal-era anchor block can be reorged under a light client, so the runtime recomputes a different implicit and rejects the signature (`InvalidTransaction::BadProof`). An immortal era uses the stable genesis hash, eliminating the race.

PR #705, opened 2026-07-27
ci: let dependabot security updates through for cargo

x3c41a:
The cargo ignore-all rule was meant to stop version bumps but it also breaks security updates: dependabot fires a job per advisory, finds every version ignored, and fails (16 failed runs since Jul 24, all_versions_ignored). Replaced with open-pull-requests-limit: 0, which stops version PRs but lets security lockfile bumps through.


.github/dependabot.yml
```diff
@@ -4,8 +4,10 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
-    ignore:
-      - dependency-name: "*"
+    # No version-update PRs (deps track the pinned polkadot-sdk), but let
```
bkontur, line 7: @x3c41a what does this mean exactly " No version-update PRs (deps track the pinned polkadot-sdk)"? That dependabot will bump all other crates except PolakdotSDK's ones? What if we release PolkadotSDK patches?

x3c41a: valid point but the previous `ignore-all` had the same problem (e.g. #662 was manual)

before: no PRs at all, vulnerability-fix jobs failing.
after: still no version PRs, but known vulnerabilities get fix PRs

PR #708, opened 2026-07-28
sdk: retry preimage-auth store once on Invalid/Payment

x3c41a:
Retries the unsigned preimage-authorized store once when polkadot-api settles it as Invalid/Payment after a reorg retracts the just-created authorization, and waits for finality on the authorization in the preimage-store integration test.

### DIFF
--- a/sdk/typescript/src/async-client.ts
+++ b/sdk/typescript/src/async-client.ts
@@ -254,6 +254,15 @@ function isAncientBirthBlockError(err: unknown): boolean {
   return e?.type === "Invalid" && e?.value?.type === "AncientBirthBlock"
 }
 
+// Same shape-matching as isAncientBirthBlockError; only Invalid/Payment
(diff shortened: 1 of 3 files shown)


RafalMirowski1 (commented):
would it not be enough that we make sure we call 
```
authorizePreimage(...).withWaitFor("finalized")
```
before we do any stores? Or is the retry needed in that case too?

dr333ws:
> would it not be enough that we make sure we call
> 
> ```
> authorizePreimage(...).withWaitFor("finalized")
> ```
> 
> before we do any stores? Or is the retry needed in that case too?

From test-side perspective - yes, but from user-side - no. Because the default is `in_block` and `authorize-then-store` is the regular flow where PAPI settles the store as permanently invalid after the reorg, so users need the retry

RafalMirowski1 (approved):

PR #715, opened 2026-07-29
Extract the split-independent changes from #571

bkontur:
### What

Extracts the fixes and refactors that ride along in #571 (the renewal pallet split) but do not depend on it, so they can land on their own and #571 shrinks to the split itself.

**Pool params.** One runtime-supplied `ValidTransactionParams { tag_prefix, priority, longevity }` per call family, replacing the `StoreRenew{Priority,Longevity}` and `RemoveExpiredAuthorization{Priority,Longevity}` pairs and the `ValidTransaction` builder chain spelled out at each construction site. Prefixes must be pairwise distinct, which `integrity_test` now checks for every family — including `authorize_*`, whose prefix is unread today (it validates untagged) but would otherwise be an exception the check had to omit.

**A pool conflict fixed.** Preimage-authorized `store` and `force_renew` both tagged `("TransactionStorageStoreRenew", content_hash)`, so a renew evicted a pending store of the same blob. They now sit on distinct prefixes, as the signed path already did.

**hop-promotion.** `PromoteTxParams` comes from the runtime, and `integrity_test` asserts `promote < store` priority directly instead of a test that stored, renewed and compared emitted priorities. `authorize_promote` also reported a bad blob size as `InvalidTransaction::Custom(0)` — which is `IMPOSSIBLE`, "implies a bug" — so an empty or oversized HOP blob told the submitter the node had hit an internal error. It reports `BAD_DATA_SIZE` now; two tests had pinned the wrong value.

**try-state direction.** `do_try_state` errored on any drift between `PermanentStorageUsed` and `Σ renewed + Σ prepaid`. Both directions are reachable on live state through history this pallet did not create — pre-counter `Renew` entries were never charged (under-count), and Root dropping a prepaid registration never decrements (over-count) — so it fails on real chains and would fire in the check-migration jobs. The equality is now enforced under `cfg(test)`, where state is built from nothing, and logged on a live chain. `used <= MaxPermanentStorageSize` stays a hard error. The drift test covered only the over-count direction; it now covers both.

**Docs.** `cargo doc --no-deps` went from ten warnings to zero.

**Tests and CI.** The pruning test ran the full `NODE_LOG_CONFIG` firehose, whose libp2p/sync trace targets truncated the log files before the failing test ran; `parachain_auto_renew_vs_no_renew_eviction_test` anchored its finality wait to the wrong blob and raced under slower finalization; zombienet log artifacts carried the RocksDB `db/` noise; two clippy nits in stress-test.

### Behaviour vs `main`

No consensus change. Nothing touches `pallet::storage`, `pallet::call_index`, `pallet::weight`, `pallet::event`, `pallet::error` or any dispatch body; `extension.rs`, `benchmarking.rs` and `weights.rs` are untouched. Every functional hunk is in pool validation, `integrity_test`, or `do_try_state`. Priorities and longevities are byte-identical to `main` (store/renew/authorize `MAX/4`, cleanup `MAX`, all `DAYS`).

Four observable differences:

1. Pool tag prefixes on the preimage path change — the deliberate fix above. Pool tags are node-local, not consensus, so no upgrade coordination.
2. `promote`'s rejection code for a bad blob size goes from `Custom(0)` to `Custom(1)`.
3. The four old pool constants are renamed and reshaped in metadata (see below).
4. try-runtime `try_state` warns instead of erroring on counter drift — off-chain tooling only, and a deliberate loosening.

`integrity_test` gains a new panic path for a mis-wired runtime (shared prefix, or `promote` priority not below `store`). Both runtimes pass, as the `construct_runtime`-generated integrity test covers.

### Metadata

All seven pool-param items are `#[pallet::constant]`, so pool pricing stays readable from metadata. (`#[pallet::constant]` needs only `TypeInfo + Encode`, not `Decode` or `MaxEncodedLen`, and `&'static str` has both via blanket impls.) The old names are still gone, though — a consumer reading `StoreRenewPriority` breaks and must read `StoreTxParams` instead, whose value encodes as compact-length prefix, prefix bytes, then two `u64`s.

### Left in #571

Split scaffolding: the `LeafValidator` / `ValidateAuthorizedCalls` rework, `TransactionInfo<Meta>` / `AuthorizationExtent<Extra>` genericization, the relocation migration, the regenerated `sdk/metadata.scale`, and the `DataRenewal` fallbacks in console-ui and the SDKs.

### Release impact

Breaking for `pallet-bulletin-transaction-storage` and `pallet-bulletin-hop-promotion` (`Config` changes). `bulletin-transaction-storage-primitives` is purely additive, so a caret-compatible bump keeps `pallet-bulletin-transaction-storage-runtime-api` and `bulletin-pallets-common` out of the release.

### Known, not addressed here

The try-state equality only becomes unconditional once its two drift causes are retired: `disable_auto_renew`'s Root arm drops a `paid` registration without decrementing `PermanentStorageUsed`, and legacy pre-counter `Renew` entries need a migration that recomputes the counter. Both are consensus changes, out of scope for a port. Separately, the nine `parachain_sync_storage.rs` call sites still use the truncating `NODE_LOG_CONFIG`.

### DIFF
--- a/pallets/hop-promotion/src/lib.rs
+++ b/pallets/hop-promotion/src/lib.rs
@@ -94,6 +94,7 @@ pub mod pallet {
 		type WeightInfo: crate::WeightInfo;
 
 		/// Pool params for `promote`. `integrity_test` enforces its priority below `store`'s.
+		#[pallet::constant]
 		type PromoteTxParams: Get<ValidTransactionParams>;
 	}
 
--- a/pallets/hop-promotion/src/lib.rs
+++ b/pallets/hop-promotion/src/lib.rs
@@ -70,9 +70,7 @@ pub mod pallet {
 	};
 	use frame_support::pallet_prelude::*;
 	use frame_system::pallet_prelude::*;
-	use pallet_bulletin_transaction_storage::{
-		Config as TxStorageConfig, WeightInfo as _, BAD_DATA_SIZE,
-	};
+	use pallet_bulletin_transaction_storage::{WeightInfo as _, BAD_DATA_SIZE};
 	use sp_runtime::{
 		traits::{IdentifyAccount, Verify},
 		AccountId32, MultiSignature, MultiSigner,
@@ -104,7 +102,7 @@ pub mod pallet {
 		fn integrity_test() {
 			// Promotion must not outbid the `store` traffic it shares blockspace with.
 			let promote = T::PromoteTxParams::get().priority;
-			let store = <T as TxStorageConfig>::StoreTxParams::get().priority;
+			let store = T::StoreTxParams::get().priority;
 			assert!(
 				promote < store,
 				"promote priority ({promote}) must be strictly below store priority ({store})",
--- a/pallets/hop-promotion/src/lib.rs
+++ b/pallets/hop-promotion/src/lib.rs
@@ -70,7 +70,9 @@ pub mod pallet {
 	};
 	use frame_support::pallet_prelude::*;
 	use frame_system::pallet_prelude::*;
-	use pallet_bulletin_transaction_storage::{WeightInfo as _, BAD_DATA_SIZE};
+	use pallet_bulletin_transaction_storage::{
+		Config as TxStorageConfig, WeightInfo as _, BAD_DATA_SIZE,
+	};
 	use sp_runtime::{
 		traits::{IdentifyAccount, Verify},
 		AccountId32, MultiSignature, MultiSigner,
@@ -93,9 +95,7 @@ pub mod pallet {
 		/// Weight information for this pallet.
 		type WeightInfo: crate::WeightInfo;
 
-		/// Pool params for `promote`. Its priority must stay below the `store` priority —
-		/// promotion only fills otherwise-unused blockspace — which `integrity_test`
-		/// enforces.
+		/// Pool params for `promote`. `integrity_test` enforces its priority below `store`'s.
 		type PromoteTxParams: Get<ValidTransactionParams>;
 	}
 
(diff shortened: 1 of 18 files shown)


pallets/hop-promotion/src/lib.rs
```diff
@@ -135,7 +161,7 @@ pub mod pallet {
 				return Err(InvalidTransaction::Call.into());
 			}
 			if !pallet_bulletin_transaction_storage::Pallet::<T>::data_size_ok(data.len()) {
-				return Err(InvalidTransaction::Custom(0).into());
+				return Err(BAD_DATA_SIZE.into());
```
bkontur, line 161: @franciscoaguirre is this ok? or should we keep 0?

franciscoaguirre: Looks good

pallets/hop-promotion/src/mock.rs
```diff
@@ -77,10 +80,22 @@ impl pallet_timestamp::Config for Test {
...
 	pub const AuthorizationPeriod: BlockNumberFor<Test> = 10;
-	pub const StoreRenewPriority: TransactionPriority = TransactionPriority::MAX;
-	pub const StoreRenewLongevity: TransactionLongevity = 10;
-	pub const RemoveExpiredAuthorizationPriority: TransactionPriority = TransactionPriority::MAX;
-	pub const RemoveExpiredAuthorizationLongevity: TransactionLongevity = 10;
+	// `static` so a test can lower the priority and check `integrity_test` rejects it.
+	pub static StoreTxParams: ValidTransactionParams =
+		ValidTransactionParams::new("Store", PRIORITY, LONGEVITY);
+	pub const PromoteTxParams: ValidTransactionParams =
+		ValidTransactionParams::new("HopPromotion", 0, 5);
```
franciscoaguirre, line 87: Let's move this into constants as well

runtimes/bulletin-paseo/src/storage.rs
```diff
@@ -111,9 +132,13 @@ parameter_types! {
 	/// Maximum allowable skew between the user's submit timestamp and the on-chain
 	/// time when validating a HOP promotion: 48 hours, in milliseconds.
 	pub const SubmitTimestampTolerance: u64 = 48 * 60 * 60 * 1000;
+	// Lowest priority: promotion only fills blockspace stores would not have used.
+	pub const PromoteTxParams: ValidTransactionParams =
+		ValidTransactionParams::new("HopPromotion", 0, 5);
```
franciscoaguirre, line 137: Same

runtimes/bulletin-westend/src/storage.rs
```diff
@@ -139,9 +160,13 @@ parameter_types! {
 	/// Maximum allowable skew between the user's submit timestamp and the on-chain
 	/// time when validating a HOP promotion: 48 hours, in milliseconds.
 	pub const SubmitTimestampTolerance: u64 = 48 * 60 * 60 * 1000;
+	// Lowest priority: promotion only fills blockspace stores would not have used.
+	pub const PromoteTxParams: ValidTransactionParams =
+		ValidTransactionParams::new("HopPromotion", 0, 5);
```
franciscoaguirre, line 165: Same

rosarp (approved):

pallets/hop-promotion/src/mock.rs
bkontur: > Let's move this into constants as well

Actually, I moved this up and down couple of times, I had it as const, then not then again, I think let's keep it on runtime-level to configure

PR #721, opened 2026-07-30
transaction-storage: one owner for the block-entry append

bkontur:
**Why:** in #571 the renewal logic moves into its own pallet, and it must not use this pallet's storage items directly. `BlockTransactions` and `TransactionByContentHash` stay `pub(super)`; the renewal pallet appends through the handle instead, so the `BlockTransactions` / `TransactionByContentHash` / `MaxBlockTransactions` / host-index invariants stay in the pallet that owns them. Extracted here so #571 does not have to carry it.

**What:** the store and renew paths each derived the cumulative `block_chunks`, pushed to `BlockTransactions` and repointed `TransactionByContentHash`. `BlockTransactionsMut` now owns that, and `with_block_transactions` is the single place that opens `BlockTransactions` — one read/write per closure, so the auto-renew drain still amortizes a single access across the batch.

`push` appends without indexing and is private, so both paths go through `store` or `renew`, which own the `transaction_index::{index,renew}` host call — an entry can no longer be appended without its data being indexed. `do_renew_in_memory` and `append_to_block_transactions` are gone.

No behaviour change: same storage operations in the same order, host call still after the runtime mutation.


pallets/transaction-storage/src/lib.rs
rosarp: stale reference to removed `fn do_renew_in_memory`

```suggestion
		/// an in-memory accumulator through repeated [`BlockTransactionsMut::renew`] calls.
```

rosarp: stale reference to removed `fn do_renew_in_memory`


```suggestion
		/// - [`BlockTransactionsMut::renew`] returns `None` because the per-block transaction slot cap
```

pallets/transaction-storage/src/lib.rs
```diff
@@ -1515,7 +1491,7 @@ pub mod pallet {
 		/// Common implementation for [`store`](Self::store) and
 		/// [`store_with_cid_config`](Self::store_with_cid_config).
 		///
-		/// FOOTGUN: `sp_io::transaction_index::index` (called below) indexes the
+		/// FOOTGUN: `sp_io::transaction_index::index` (via `Self::store_entry`) indexes the
```
rosarp:
```suggestion
		/// FOOTGUN: `sp_io::transaction_index::index` (via `BlockTransactionsMut::store`) indexes the
```

pallets/transaction-storage/src/lib.rs
```diff
@@ -2526,6 +2457,50 @@ pub mod pallet {
...
+		pub(crate) fn store(&mut self, entry: TransactionInfo) -> Option<u32> {
+			let (extrinsic_index, size, content_hash) =
+				(entry.extrinsic_index, entry.size, entry.content_hash);
+			let new_index = self.push(entry)?;
+			sp_io::transaction_index::index(extrinsic_index, size, content_hash);
+			Some(new_index)
+		}
+
+		/// `Self::store` for already-stored data: appends a `Renew` copy of `info` under
+		/// `extrinsic_index` and re-indexes it with the host. Same no-rollback caveat.
```
rosarp: Adding hard-cap accounting check similar to `fn do_renew`.

```suggestion
        /// [`Self::store`] for already-stored data: appends a `Renew` copy of `info` under
        /// `extrinsic_index` and re-indexes it with the host. Same no-rollback caveat.
        ///
        /// Appends only — the caller owns the economics. This does not check that
        /// `info.content_hash` is still stored, and does not charge the renewal: the
        /// per-account `bytes_permanent` debit and the chain-wide [`PermanentStorageUsed`]
        /// cap check must already have run for `info.size` (see
        /// [`Pallet::check_authorization`]). Skipping them grants permanent retention for
        /// free and leaves the counter out of step with the `Renew` entries it accounts for.
```

rosarp (approved):

PR #738, opened 2026-08-10
Renewals follow-ups 2

rosarp:


sdk/rust/src/transaction.rs
```diff
@@ -291,7 +331,7 @@ impl TransactionClient {
 
 		let Some(bytes) = maybe_auth else { return Ok(None) };
 
-		let auth = StoredAuthorization::decode(&mut &bytes[..])
+		let auth = StoredAuthorization::decode_any(&bytes)
```
bkontur: I am almost sure that this hacky decoding and Direct reading storage item, should be replaced by versioned runtime api call.

bkontur:
@rosarp please, when https://github.com/paritytech/polkadot-bulletin-chain/pull/739 is merged, rebase this one

docs/todo.html
karolk91, line 1: do we need this todo file ?

rosarp: No we don't. That was my local tracker. Sorry, will remove it

sdk/rust/src/transaction.rs
rosarp: Fixed this.

rosarp:
@karolk91 @bkontur Anything needed in this PR? 
It was supposed to go after renewals migrations changes.

sdk/rust/src/types.rs
```diff
@@ -535,6 +554,7 @@ mod tests {
 			"CID_CALCULATION_FAILED",
 			"TRANSACTION_FAILED",
 			"INVALID_CHUNK_SIZE",
+			"UNSUPPORTED_OPERATION",
 		];
```
karolk91:
```suggestion
assert_eq!(all_errors().len(), expected.len(), "expected[] out of sync with all_errors()");
```

.github/workflows/integration-test.yml
```diff
@@ -163,8 +163,9 @@ jobs:
 
       # Chopsticks cannot synthesize custom pallet inherents
       # (AcalaNetwork/chopsticks#1037), so the drain inherent is injected by hand.
-      # Skipped for external runtimes: those are already-deployed runtimes that
-      # predate `pallet-bulletin-data-renewal`.
+      # Skipped for external runtimes: none of them wire
```
karolk91: Since polkadot runtime will contain the pallet, we should prepare for that and avoid skipping for all external runtimes

rosarp: I have added a flag in runtime_matrix for data_renewal
@karolk91 check once this fix

karolk91 (approved):

PR #740, opened 2026-08-11
Compose the account_authorization runtime API in the runtime

bkontur:
`account_authorization` lived in `pallet-bulletin-data-renewal` only because its
`bytes_permanent_used` field reads `PermanentExtent`. The runtime is the only place that
names both the storage pallet's opaque `AuthorizationExtra` and the runtime-API struct, so
the summary is composed there now — no new trait, no `Config` bound, and the runtime-API
struct is unchanged.

This drops the `pallet-bulletin-transaction-storage-runtime-api` dependency from
`pallet-bulletin-data-renewal`.

Plus three small cleanups: name the decoded value in `transactions_at` `txs` instead of `v3`,
fold the single-test `tests/runtime_api.rs` back into `tests.rs`, and call the bulletin
pallets through their runtime aliases (`TransactionStorage`, `DataRenewal`, `HopPromotion`).


karolk91 (approved):

pallets/data-renewal/src/tests.rs
```diff
@@ -2985,25 +2985,31 @@ fn remove_expired_account_authorization_succeeds_with_outstanding_renewals() {
...
+/// `get_active_authorization`; this pins what they read — the extent, the expiration,
+/// and this pallet's renew quota in `extra` — including that expiry hides the entry.
 #[test]
-fn account_authorization_returns_none_when_missing_or_expired() {
+fn active_authorization_carries_the_runtime_api_summary_fields() {
 	new_test_ext().execute_with(|| {
 		run_to_block(1, || None);
 		let who = 1;
+		let scope = AuthorizationScope::Account(who);
+		let active = || txs::Pallet::<Test>::get_active_authorization(&scope);
```
karolk91: I think it should work like below (didnt check)
```suggestion
		let active = || TransactionStorage::get_active_authorization(&scope);
```

bkontur: > I think it should work like below (didnt check)

thank you, right, sorry, before I was quick to apply suggestion, my assistant pushed it

rosarp (approved):

x3c41a (approved):
Looks good

PR #759, opened 2026-08-19
ci: keep external runtime checkouts out of target/

rosarp:
Causing CI to fail due to cache miss
|Issue | CI Build failure|
| --- | --- |
| https://github.com/paritytech/polkadot-bulletin-chain/pull/738 | / https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/32051138348/job/95451163257?pr=738 |
| https://github.com/paritytech/polkadot-bulletin-chain/pull/753 | / https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/32251895470/job/96064992991?pr=753 }

## What is broken

The `bulletin-polkadot` integration job fails in its first step, "Build external runtime", after about a minute. It never gets to compiling our code or running tests:

🔨 Building bulletin-polkadot-runtime [features=default]...
cd: .../target/external-runtimes/bulletin-polkadot/./system-parachains/bulletin/bulletin-polkadot: No such file or directory

This is not specific to a branch, `main` fails the same way. The pattern is the cache:

| run | branch | cache | result |
|---|---|---|---|
| 32246772062 | main, 11:16Z | restored | pass |
| 32251079129 | main, 12:08Z | `Cache not found` | fail |
| 32251895470 | rs-batch-perm-storage-used-event | `Cache not found` | fail |
| 32261255386 | rs-rename-pendingAutoRenewals | `Cache not found` | fail |

## Why it happens

The checkout was placed under `target/`, and `Swatinem/rust-cache` also owns that directory. When it restores `target/`, we can end up with a `.git` but no files next to it. Our recipe then skips cloning, because it clones only when `.git` is missing, and `git checkout FETCH_HEAD` does nothing, because HEAD already points at the pinned commit. So no file is restored, no command fails, and only the `cd` at the end complains. I reproduced this locally by deleting the files while keeping HEAD at the pinned commit: a normal checkout leaves the directory missing, a forced checkout brings it back.

There is a second problem behind the first one. We cached the whole checkout together with its `target/`, which is a few GB per entry. The repo has 10 GB for all caches, so these entries push each other out. That is most likely why the cache was missing in the first place.

## What this PR changes

**Cache only the WASM, around 2 MB, and nothing else.** The checkout and its `target/` are not worth keeping between runs, and once we stop caching them, the broken checkout above cannot happen anymore. A cache hit is now also cheaper, because we do not touch git at all. There is no downside for build time: the key is the commit sha, so a miss always meant a full build.

| path | cached | what it is |
|---|---|---|
| `.external-runtimes/src/<name><variant>` | no | checkout we build in |
| `.external-runtimes/wasm/<name><variant>/<sha>/<pkg>.compact.compressed.wasm` | yes | the blob |

**Keep both outside `target/`,** so that no other cache action can touch them.

**Use `git checkout --force`.** CI cannot produce a half restored checkout anymore, but a local one can still be in a strange state, and a forced checkout brings it back to the pinned commit.

**Check that the runtime directory is really there before building.** If it is missing we clone once more, and if it is still missing we stop with a clear message instead of a confusing `cd` error. This is also the message you would get if the external repo moves the runtime to another directory, and it points to the matrix entry that needs an update.

**Compute the sha, the paths and the cache key in one place.** `just external-runtime-paths <name> [features]` prints them and the action reads it, so the cache path and the recipe can no longer disagree. Until now the action and the justfile both resolved the ref to a sha, both built the variant suffix, both spelled out the cache path, and the workflows passed `repo` and `ref` that the recipe read from the matrix anyway. The key includes the variant, so the normal and the `try-runtime` build of the same runtime do not share one entry:

external-runtime-wasm-bulletin-polkadot-e93abddf…
external-runtime-wasm-bulletin-polkadot-try-runtime-e93abddf…

## Notes

- The blob path now contains the sha. Before, a new pin was checked out but the old blob was reused, which was wrong.
- Local reruns work offline now, and the checkout keeps its `target/`, so dependencies are still reused after a pin bump.
- If you edit the checkout under `.external-runtimes/src/` by hand, the forced checkout will throw those edits away. For testing a modified external runtime, point `external_runtime.repo`/`ref` in `scripts/runtimes-matrix.json` to your own commit.
- The first run after merge builds the external runtime once per runtime and feature variant, as with any change of a cache key.

### DIFF
--- a/examples/justfile
+++ b/examples/justfile
@@ -134,7 +134,7 @@ bulletin-parachain-zombienet-start test_dir runtime="bulletin-westend-runtime":
     just _wait-for-rpc $ZOMBIENET_PID "{{ test_dir }}/zombienet.log"
 
 # Build a runtime that lives in an external (Fellows) repo and print the WASM path.
-# Different feature sets use separate checkouts so their build caches don't collide.
+# Different feature sets get separate checkouts and cached blobs.
 build-external-runtime name features="":
     #!/usr/bin/env bash
     set -e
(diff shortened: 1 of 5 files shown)


karolk91:
@rosarp while I think the main idea/problem here still applies (the target folder being an issue). Since you're onto some refactor, I think we should consider caching only the resulting WASM blobs, avoid caching repos/targets since it costs a lot of limited cache available in the CIs

rosarp:
@karolk91 Thanks! Didn't realize but this is super!
Cache size from GBs to MBs now. Fixed it.

examples/justfile
```diff
@@ -152,22 +152,60 @@ build-external-runtime name features="":
...
+        if [ -z "$SHA" ]; then
+            >&2 echo "❌ Cannot resolve $REF in $REPO"
+            exit 1
+        fi
+    fi
+
+    # Only the blob is worth keeping between runs: the checkout and its `target/` run to
+    # several GB, enough to evict every other cache in the repo. Keyed by sha so a pin
+    # bump cannot serve a stale blob.
+    WASM_DIR="$ROOT_DIR/.external-runtimes/wasm/{{ name }}${VARIANT_SUFFIX}/$SHA"
```
karolk91: I think this logic now lives in two places here and in the action.yml, lets extract it to separate justfile target like
```
external-runtime-wasm-dir name features="":
      #!/usr/bin/env bash
      set -e
      …read repo+ref from runtimes-matrix.json, compute suffix, resolve sha…
      EXT_ROOT="${EXTERNAL_RUNTIMES_DIR:-$ROOT_DIR/.external-runtimes}"
      echo "${EXT_ROOT}/wasm/{{ name }}${SUFFIX}/$SHA"
```

and use in both places
```suggestion
    WASM_DIR="$ROOT_DIR/$(just external-runtime-wasm-dir '{{ name }}' '{{ features }}')"
```

rosarp: Found one issue in runtime paths where a cache hit restores nothing.
Overall, removed duplicates and consolidated in one way path generation.

karolk91 (approved):
Looks good, thank you!

PR #762, opened 2026-08-23
ci: package all crates in one cargo invocation for dry run

karolk91:
The `Publish Crates` dry run has never been able to pass for an unpublished version. [Run 32664551472](https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/32664551472) (dry run of 0.12.0) failed on the third crate:
```
  Packaging pallet-bulletin-transaction-storage-runtime-api v0.12.0
      Updating crates.io index
  error: failed to prepare local package for uploading
  Caused by:
    failed to select a version for the requirement bulletin-transaction-storage-primitives = "^0.12.0"
    candidate versions found which didn't match: 0.11.0, 0.10.0-draft, 0.9.0-draft, ...
```
`Update crate versions` pins every intra-repo dep to the new version, and `cargo package` strips the `path` to resolve that dep from the registry when it builds the packaged `Cargo.lock`. The new version isn't on crates.io yet, so resolution fails. That happens while *preparing* the archive, before any verification build — so the `--no-verify` flags (and the comment explaining them) did not address the actual failure. The first two crates only passed because they have no intra-repo dependencies.

Packaging all six in a single invocation makes cargo resolve the bulletin-* deps against the locally packaged copies instead of crates.io. Verification then works too, so `--no-verify` is dropped and the dry run becomes a real check rather than a manifest smoke test.


karolk91:
Tested dry-run in https://github.com/paritytech/polkadot-bulletin-chain/actions/runs/32665804663

rosarp (approved):

PR #768, opened 2026-08-24
Fix xcm weights: batch_call + worst-case alias_origin

karolk91:
Ports the 2606-2 XCM benchmark-helper fixes to bulletin-paseo and bulletin-westend:

- batch_call: override the default so the `Transact` benchmark weighs a real
  `Utility::batch` and recurses over nested calls (otherwise `Transact` is
  under-weighed).
- alias_origin: exercise the worst case via
  `parachains_common::benchmarking::set_up_worst_case_authorized_alias` (the
  `AuthorizedAliasers` entry added in 2606-2) instead of a hand-rolled pair.


karolk91:
/cmd bench --pallet pallet_xcm pallet_xcm_benchmarks::generic

rosarp (approved):

PR #776, opened 2026-09-01
sdk: add enable_auto_renew / disable_auto_renew to both SDKs

mudigal:
## Summary

The SDKs expose `renew` and `force_renew` but not the auto-renewal pair. This adds `enable_auto_renew` / `disable_auto_renew` to both SDKs, completing coverage of the `DataRenewal` pallet's user-facing calls.

**Rust** (`TransactionClient`):
- `enable_auto_renew(content_hash, signer, wait_for)` and `disable_auto_renew(content_hash, signer, wait_for)`, submitting on the `DataRenewal` pallet behind the existing `ensure_renewal_available` guard (`Error::RenewalUnavailable` on chains without the pallet).

**TypeScript** (`AsyncBulletinClient`):
- `enableAutoRenew(contentHash)` / `disableAutoRenew(contentHash)` CallBuilders, taking the 32-byte content hash and encoding it as PAPI SizedHex.
- The cached renewal resolution now also reports whether the resolved pallet carries the auto-renew pair — compat levels via `getStaticApis()` on real clients, entry presence on mock/hand-rolled api objects — so both methods reject with `UNSUPPORTED_OPERATION` on runtimes that lack the calls (including pre-`TransactionRef` legacy runtimes).
- `MockBulletinClient` records the new operations.

READMEs updated to list the full renewal surface.

## Testing

- `cargo check` / `cargo clippy -p bulletin-sdk-rust --all-targets --all-features` clean; `cargo test -p bulletin-sdk-rust` passes.
- TS: `tsc --noEmit`, biome, and `npm run test:unit` (116 tests) pass; new renew-shape tests pin auto-renew resolution and argument encoding across DataRenewal / pre-split / legacy runtimes.

Companion docs PR updates the SDK book renewal pages.

### DIFF
--- a/sdk/typescript/src/async-client.ts
+++ b/sdk/typescript/src/async-client.ts
@@ -120,10 +120,19 @@ type RenewalPallet = {
     args: { block: number; index: number } | { entry: TransactionRef },
   ): PapiTransaction
   force_renew?(args: { entry: TransactionRef }): PapiTransaction
+  enable_auto_renew?(args: { content_hash: string }): PapiTransaction
+  disable_auto_renew?(args: { content_hash: string }): PapiTransaction
 }
 
-/** Where the runtime's renewal extrinsics live and which call shape they take. */
-type RenewalResolution = { pallet: RenewalPallet; shape: RenewShape }
+/**
+ * Where the runtime's renewal extrinsics live, which call shape they take,
+ * and whether it carries the auto-renewal pair.
+ */
+type RenewalResolution = {
+  pallet: RenewalPallet
+  shape: RenewShape
+  autoRenew: boolean
+}
 
 /** The slice of PAPI's `getStaticApis()` result the renewal resolution reads. */
 type StaticApisCompat = {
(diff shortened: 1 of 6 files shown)


RafalMirowski1 (commented):
I think the typescript part of this was already covered in https://github.com/paritytech/polkadot-bulletin-chain/pull/802 so only the rust part is necessary.

sdk/typescript/src/async-client.ts
```diff
@@ -1322,11 +1333,17 @@ export class AsyncBulletinClient implements BulletinClientInterface {
...
+          return {
+            pallet: dataRenewal,
+            shape: "transactionRef",
+            autoRenew: !!dataRenewal.enable_auto_renew,
+          }
         }
+        const storagePallet = transactionStorage as RenewalPallet
         return {
-          pallet: transactionStorage,
+          pallet: storagePallet,
```
bkontur, line 1344: @RafalMirowski1 @mudigal shouldn't we remove `const storagePallet = transactionStorage as RenewalPallet` and use directly `dataRenewal`?

```suggestion
          pallet: dataRenewal,
```

PR #782, opened 2026-09-02
dashboards: add HOP metrics to Bulletin Paseo

dr333ws:
Extends the Bulletin Paseo dashboard (uid `bulletin-paseo`) with the full HOP metric set from [`sc-hop`: add Prometheus metrics (polkadot-sdk#12662)](https://github.com/paritytech/polkadot-sdk/pull/12662), so pool, promotion and RPC health sit next to the existing liveness, IPFS and bitswap panels.

These metrics started flowing on Paseo after [devops-cloud-infra#4274](https://github.com/paritytech/devops-cloud-infra/pull/4274) bumped the bulletin-next collators to the master image with HOP metrics.

Added rows:
- Pool: fill %, bytes vs max, entries, inserted-bytes rate, removals by reason, expired-unpromoted (data-loss upper bound)
- Promotion: backlog, promotions confirmed, maintenance-tick liveness
- RPC: calls, errors by reason, error ratio (benign not_found/already_claimed excluded, clamped 0 to 100%), p95 latency

Adds a `node` template variable to scope panels per collator.

Verify against live Paseo data: https://grafana.teleport.parity.io/d/bulletin-paseo-new/bulletin-paseo-new?orgId=1&from=now-6h&to=now&timezone=utc&var-datasource=PC96415006F908B67&var-chain=$__all&var-node=$__all&refresh=1m

PR #783, opened 2026-09-02
console-ui: link the Bulletin dashboard on the Ops page

dr333ws:
Follow-up to #782

Adds a `bulletin` monitoring link and a "Bulletin Dashboard" card under Chain Health, wired for westend, products-devnet, paseo-next-v2 and polkadot. The dashboard is parameterised by `var-chain`, so the one dashboard serves every network, including Bulletin Polkadot (`var-chain=bulletin-polkadot`) without a separate copy.

The card links to chain liveness, IPFS gateway, Bitswap, and the HOP pool/promotion/RPC panels added in #782.

### DIFF
--- a/console-ui/src/config/networks.ts
+++ b/console-ui/src/config/networks.ts
@@ -15,6 +15,8 @@ export interface MonitoringLinks {
   grafana?: string;
   /** Grafana Bitswap server insights dashboard (IPFS/Bitswap serve load). */
   bitswap?: string;
+  /** Grafana Bulletin dashboard: chain liveness, IPFS, Bitswap, HOP pool/promotion/RPC. */
+  bulletin?: string;
   /** Sentry dashboard for product-side telemetry on this chain. */
   sentry?: string;
   /** Sentry drill-down: deploy.storage phase (per-deploy Bulletin write). */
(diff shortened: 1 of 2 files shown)


dr333ws:
Screenshot 2026-09-03 at 12 02 49
WAI

console-ui/src/config/networks.ts
```diff
@@ -121,6 +123,12 @@ function bitswapLink(chain: string): string {
   );
 }
 
+// Bulletin dashboard (uid `bulletin-paseo`). Paseo only; it is the sole Bulletin dashboard.
+const GRAFANA_BULLETIN =
+  "https://grafana.teleport.parity.io/d/bulletin-paseo/bulletin-paseo" +
+  "?orgId=1&from=now-6h&to=now&timezone=utc" +
```
dr333ws: timezone might be redundant...

console-ui/src/config/networks.ts
```diff
@@ -121,6 +123,12 @@ function bitswapLink(chain: string): string {
   );
 }
 
+// Bulletin dashboard (uid `bulletin-paseo`). Paseo only; it is the sole Bulletin dashboard.
+const GRAFANA_BULLETIN =
+  "https://grafana.teleport.parity.io/d/bulletin-paseo/bulletin-paseo" +
```
bkontur: bulletin-paseo/bulletin-paseo? 

I think we extracted everything network-related to network.rs or some config

dr333ws: I left it intentionally because right now we only have the dashboard for Paseo (Summit network was decomissioned, Mainnet/Polkadot is unavailable atm). I don't strongly oppose but I don't see a point in abstracting it away yet if we only have one invariant...

dr333ws: or if you're asking about double `bulletin-paseo` that's by design because grafana format is `/d/<uid>/<slug>`, where `slug` is derived from the title thus it's the same as `uid`

dr333ws: though, from what I see it also works without the `slug`

console-ui/src/config/networks.ts
```diff
@@ -244,6 +252,7 @@ export const BULLETIN_NETWORKS: Record<string, Network> = {
         "next-bulletin-paseo",
         "paseo-bulletin-next-collator-node-0",
       ),
+      bulletin: GRAFANA_BULLETIN,
```
bkontur: why not just add the link directly? will we reuse it?

```suggestion
      bulletin: "https://grafana.teleport.parity.io/d/bulletin-paseo?orgId=1&from=now-6h&to=now&timezone=utc&var-datasource=PC96415006F908B67&var-chain=next-bulletin-paseo",
```

if we want to keep, I suggest GRAFANA_BULLETIN -> GRAFANA_BULLETIN_PASEO_NEXT_V2

dr333ws: done

bkontur (approved):

PR #789, opened 2026-09-07
Add pallet tests for live RetentionPeriod changes

### LINKED ISSUE #159: [Investigate/Testing] Add test-case to verify that changing `RetentionPeriod` (+/-) live does not break anything
Now we have `RetentionPeriod` as storage item exposed by runtime API, so the node binaries are reading it directly, no hard-coded values anymore.

We should have some kind of test (or at least verify), probably on the runtime level or maybe pallet level is enough?

Something like:

```
RetentionPeriod = 3
Block1
- store data
Block2
- store data
Block3
- store data
Block4
- check_proof(Block1) inherent
- store data
- change RetentionPeriod = 2 (decrease should be fine)
Block5
- check_proof(Block3) inherent
- store data
- change RetentionPeriod = 4 (increase, potential problems?)
Block6
- check_proof(Block2) inherent (Block2 transaction metadata where removed ??)
  - we could have a problem here, because transaction metdata were already removed?, do we need here some more fixing - something like to store previous retention period from some block and do some magic?) 
```

The problem could be also when we just do a  decrease, there could exist potentially some gap between, which is never removed from transaction metadata. (probably not a big issue, we can cleanup manually when decreasing)

Tyagiquamar:
## Summary

- Add pallet-level tests covering decreasing and increasing \RetentionPeriod\ while the chain is storing data and checking proofs.
- Document current behavior: after a shorter retention window prunes metadata, increasing the period again cannot resurrect proofs for already-pruned blocks (\MissingStateData\).
- Verify that decreasing retention eventually prunes stale metadata and \do_try_state\ stays clean.

Fixes #159.

## Investigation result

No consensus bug found in the exercised sequence. Proof verification always targets \current_block - current_retention_period\ and requires live \Transactions[target]\ metadata.

## Tests

\\\
cargo test -p pallet-bulletin-transaction-storage retention_period
# 3 passed (includes 2 new tests)
\\\

### DIFF
--- a/pallets/transaction-storage/src/tests.rs
+++ b/pallets/transaction-storage/src/tests.rs
@@ -29,7 +29,7 @@ use super::{
 	pallet::Origin,
 	AllowedAuthorizers, AuthorizationExtent, AuthorizationOrigin, AuthorizationScope,
 	AuthorizedCaller, AuthorizerBudget, EnsureAllowedAuthorizers, Event, Quota, TransactionInfo,
-	AUTHORIZATION_NOT_EXHAUSTED, AUTHORIZATION_NOT_EXPIRED, AUTHORIZER_NOT_FOUND, BAD_DATA_SIZE,
+	TransactionStorageProof, AUTHORIZATION_NOT_EXHAUSTED, AUTHORIZATION_NOT_EXPIRED, AUTHORIZER_NOT_FOUND, BAD_DATA_SIZE,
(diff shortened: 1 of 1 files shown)


Tyagiquamar:
@bkontur when you have a chance, could you take a look at this? It addresses #159 by adding pallet tests for increasing and decreasing RetentionPeriod on a live chain to verify storage pruning and proof safety. Happy to adjust anything needed.

Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

Issue #793, opened 2026-09-10
Tune console-ui - CID panel


bkontur:
<img width="401" height="315" alt="Image" src="https://github.com/user-attachments/assets/0a851bb4-0fda-4d95-b927-45ffe772bef2" />

Hints:
- [x] Codec - display hex version + translation to human-readable (Plain / 0x55) - we should have some enum translation already
- [x] Hash - this is Hashing enum, also the same (Blake2b / 0x55) - also we have a enum with three values for this
- [x] Digest Size - ? what is this ? I think it is "Hash Size" - @karolk91 wdyt?
- [x] ContentHash - extract the plain contentHash from CID in hex format: 0x23A... 
- [x] Add CID panel to the "Renew Storage" tab
- [x] On-chain Status pannel - adjust "Storage and retention info from the chain" that it is info about the latest uploaded/renewed 
- [x] Renew panel -> "Renew selected" now calls forceRenew, which is not what we want
  - [x] "Renew selected" -> it needs to trigger `fn renew` (with some explanation that it is one-time operation, that will be renewed at the end of retention period + display this `Retention
~13d 23h left` <- we have this info here "On-chain Status")
- [x] "Renew immediatelly" -> triggers forceRenew with simple explanation
- [x] "Enable auto-renew" -> simple explanation
- [x] Those above are minimalistic version, more changes are https://github.com/paritytech/polkadot-bulletin-chain/issues/524 (can be follow-up)

PR #796, opened 2026-09-10
console-ui: name CID codec/hash and share the CID panel

rosarp:
Addresses partial concerns in : https://github.com/paritytech/polkadot-bulletin-chain/issues/793

### DIFF
--- a/console-ui/src/components/CidInfoCard.tsx
+++ b/console-ui/src/components/CidInfoCard.tsx
@@ -1,0 +1,85 @@
+// Copyright (C) Parity Technologies (UK) Ltd.
+// SPDX-License-Identifier: GPL-3.0-only
+
+import { useState } from "react";
+import { Check, Copy } from "lucide-react";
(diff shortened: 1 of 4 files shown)


rosarp:
image 
image
image


- [x] Codec - display hex version + translation to human-readable (Plain / 0x55) - we should have some enum translation already
- [x] Hash - this is Hashing enum, also the same (Blake2b / 0x55) - also we have a enum with three values for this
- [x] Digest Size - ? what is this ? I think it is "Hash Size"
- [x] ContentHash - extract the plain contentHash from CID in hex format: 0x23A...
- [x] Add CID panel to the "Renew Storage" tab 
- [x] On-chain Status pannel - adjust "Storage and retention info from the chain" that it is info about the latest uploaded/renewed

karolk91 (approved):

bkontur:
@rosarp very nice, thank you :)

PR #801, opened 2026-09-11
stress-test: add verify-all, verify-dag, ipfs-gateway

karolk91:
Three new read scenarios and an HTTP gateway transport for the stress test tool.

**`bitswap verify-all`** scans the whole `TransactionStorage::Transactions` map at the finalized block, downloads every CID exactly once, recomputes the multihash over the returned bytes and compares it with the CID. Every failing CID is listed at the end with its block number and reason. The result's `data_verified` is true only if every item downloaded and verified.

**`bitswap verify-dag`** does the same for every dag-pb root on chain, then parses each manifest client-side and follows every child link. A chunked file passes only if its whole DAG is retrievable. The chain indexes roots, not leaves, so `verify-all` cannot see missing leaves.

**`dag-fanout`** uploads N random raw leaves plus a UnixFS root via `store_with_cid_config`, one fresh account per item at nonce 0, waits for all of them at the finalized block, then fetches the root from an IPFS gateway with `?format=car&dag-scope=all` and compares every returned block with the uploaded bytes. Fetching the root makes the gateway request all N leaf CIDs from the backing node at once, so `--leaves 50,100,200` exercises wantlists of those sizes.

**`--ipfs-gateway <url>`** reads over an HTTP IPFS gateway instead of Bitswap. Accepted by `bulk-read`, `verify-all`, `verify-dag` and `dag-fanout`.

**`bulk-read` changes**
- `--random-cids`: every request uses a fresh random CID, so every request misses the cache and forces provider lookups. Runs until Ctrl+C.
- `--dag-pb`: read only dag-pb roots, each as a full CAR. Gateway only.
- Discovery now scans the whole map and samples at random up to `--read-size`, so repeated runs read different CIDs. Before, it stopped at the first `--read-size` bytes of chain history and every run read the same prefix.

## Shared code

- `fetch.rs`: `ReadSource`, `Fetcher`, `FetchMode`, `connect_fetchers`. Bitswap peers are dialled concurrently. One `HttpGatewayClient` is built per distinct URL and cloned per worker.
- `cid_hash.rs`: `Hashing` enum with variant index, multihash code, runtime variant name and digest, plus `cid_v1` and `verify_block`. This table existed in five forms across six files.
- `bitswap.rs`: `MAX_WANTLIST_CIDS` is public and `fetch_blocks` rejects a longer wantlist instead of letting the server drop it silently.
- `store.rs`: `PreSignedTx` derives `Clone`.
- `dag_pb.rs`: minimal UnixFS dag-pb encoder and link decoder, with named field and wire-type constants. The encoder is hand-rolled on purpose: prost emits fields in tag order (Data before Links), which go-ipld-prime rejects as non-canonical dag-pb.

## New dependencies

`reqwest` (rustls only), `sha2`, `sha3`, `fdlimit`. The fd limit is raised at startup because high read concurrency needs hundreds


stress-test/src/http_gateway.rs
```diff
@@ -0,0 +1,160 @@
...
+use std::time::Duration;
+
+use crate::dag_pb::read_uvarint;
+
+const REQUEST_TIMEOUT: Duration = Duration::from_secs(600);
+
+fn parse_car(car: &[u8]) -> Result<Vec<(cid::Cid, Vec<u8>)>> {
+	let mut pos = 0usize;
+	let (header_len, prefix_len) = read_uvarint(&car[pos..]).map_err(anyhow::Error::msg)?;
+	pos += prefix_len + header_len as usize;
```
rosarp, line 12: use safe asthmatics (checked_add) for `pos`

PR #806, opened 2026-09-16
console-ui: refresh PAPI metadata snapshots and add drift check

RafalMirowski1:
### What's wrong
The live console UI shows `TransactionStorage pallet unavailable — Incompatible runtime entry Storage(TransactionStorage.Authorizations)` on Bulletin Polkadot and Products Devnet.

The Pages deploy runs `papi generate` from the metadata snapshots committed in `console-ui/.papi/metadata/`, not from the live chains. Those snapshots were last refreshed in August: the Polkadot one predates the `TransactionStorage` pallet, and the Products Devnet one has the pre-renewal-split `AuthorizationExtent` layout.

### Fix
- Refresh all five snapshots with `papi update` (metadata, code hash, and Previewnet's new genesis).
- Read permanent bytes from `extent.extra`; the renewal split moved `bytes_permanent` there and PAPI unwraps the single-field struct, so the Account's Storage card always showed 0.

Verified `TransactionStorage.Authorizations.getValue` against all five live networks with the regenerated descriptors.

### Added
- `papi-drift` job in `check-console-ui.yml`: runs `papi update --skip-codegen` and fails if `.papi/metadata` or `.papi/polkadot-api.json` changes, with a message pointing to the fix.
- README note on resolving a drift failure.


.github/workflows/check-console-ui.yml
```diff
@@ -48,3 +48,33 @@ jobs:
...
+
+      # The Pages deploy generates descriptors from the committed snapshots, so a
+      # stale snapshot ships a UI that papi rejects against the live runtime.
+      - name: Fail on drift
+        working-directory: console-ui
+        run: |
+          if ! git diff --exit-code -- .papi/metadata .papi/polkadot-api.json; then
+            echo "::error::Committed PAPI metadata is behind the live chains. Run 'npx papi update' in console-ui and commit .papi/metadata and .papi/polkadot-api.json."
+            exit 1
+          fi
```
karolk91:
```suggestion
```

I think the job you added is too strict, in general the descriptors should not cause problems unless we change something in an incompatible way - so a mere check if anything changed imho is too much (eg. consider this job failing on an unrelated PR because Polkadot Bulletin runtime version changed without any other changes). As long as old descriptors are compatible with our logic, we shouldn't need to update them

There is actually a tool from PAPI team to detect incompatible runtime changes, @antkve  was working on the PR to add the check into our CI: https://github.com/paritytech/polkadot-bulletin-chain/pull/661. So I think if we have 661 completed then we would be able to detect breaking changes before releaseing a runtime - and either accept braking changes (then we need to also update descriptors) or otherwise descriptors should not need to be updated, wdyt?

console-ui/README.md
```diff
@@ -101,6 +101,8 @@ npm run papi:generate
 npm run papi:update
 ```
 
+CI fails when the committed `.papi/metadata` snapshots drift from the live chains. Run `npx papi update` and commit `.papi/metadata` and `.papi/polkadot-api.json`.
+
```
karolk91:
```suggestion
```

.github/workflows/check-console-ui.yml
RafalMirowski1: I agree especially this failing on unrelated PRs would be annoying. I think if we can detect breaking runtime changes before then that's better or we could have a daily job that does the above and if it fails opens a PR with descriptor bump or something like that but I guess detecting before the issue even better we just have to remember to bump the descriptors

karolk91 (approved):

